PR #297, opened 2021-02-15
Fix for build with SUPPORT_OGG on Android

steeleduncan:
`vorbisfile` target does not exist and this breaks the build on Android with CMake.
This seems to be referring to vorbisfile.c, which is part of `libvorbisidec`.
Removing it fixes the build.


sezero:
Thanks.

Issue #233, opened 2021-02-11
Merge SDL_mixer with SDL Mixer X (my extended fork of SDL_mixer)


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

These attachments are available in the static archive:

* [1139.patch (1139.patch, text/plain, 2019-12-21 12:19:59 +0000, 716 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4123)
* [1141.patch (1141.patch, text/plain, 2019-12-21 12:20:34 +0000, 13908 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4124)
* ~~[1142.patch (1142.patch, text/plain, 2019-12-21 12:21:01 +0000, 10728 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4125)~~
* [1144.patch (1144.patch, text/plain, 2019-12-21 12:21:36 +0000, 10793 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4126)
* [1145.patch (1145.patch, text/plain, 2019-12-21 12:22:08 +0000, 25782 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4127)
* [1147.patch (1147.patch, text/plain, 2019-12-21 12:22:41 +0000, 793 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4128)
* [1152.patch (1152.patch, text/plain, 2019-12-21 12:23:08 +0000, 2634 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4129)
* [1142.patch (cleaned-up) (1142.patch, text/plain, 2019-12-21 17:10:33 +0000, 10530 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4132)
* [1145a.patch (clean-ups for 1145.patch) (1145a.patch, text/plain, 2019-12-21 17:13:22 +0000, 3866 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4133)
* [An example MP3 file with the sequence of zeroes at begin (moi_drug_luchshe_vseh_igraet_blu.mp3.7z, application/octet-stream, 2021-02-05 19:13:50 +0000, 9878829 bytes)](https://bugzilla.libsdl.org/attachment.cgi?id=4764)

**Reported in version:** unspecified
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

[...]

PR #300, opened 2021-02-18
music_mad.c: Added the support for the Tell and Duration

Wohlstand:
This update I developed a while ago for the MixerX fork. It allows reporting the song duration and the current time position while playing the song.

#233

### DIFF
--- a/src/codecs/music_mad.c
+++ b/src/codecs/music_mad.c
@@ -150,9 +150,164 @@ typedef struct {
     unsigned short last_nchannels;
     unsigned int last_samplerate;
 
+    double total_length;
+    int sample_rate;
+    int sample_position;
+
     unsigned char input_buffer[MAD_INPUT_BUFFER_SIZE + MAD_BUFFER_GUARD];
 } MAD_Music;
 
+static void read_update_buffer(struct mad_stream *stream, MAD_Music *music);
+
+static double extract_length(struct mad_header *header, struct mad_stream *stream, Sint64 file_size)
+{
+    int mpeg_version = 0;
+    int xing_offset = 0;
+    Uint32 samples_per_frame = 0;
+    Uint32 frames_count = 0;
+    unsigned char const *frames_count_raw;
+
+    /* There are two methods to compute duration:
+     * - Using Xing/Info/VBRI headers
+     * - Rely on filesize and first size of frame in condition of CRB
+     * https://www.codeproject.com/Articles/8295/MPEG-Audio-Frame-Header#VBRHeaders
+     */
+
+    if (!stream->this_frame || !stream->next_frame ||
+        stream->next_frame <= stream->this_frame ||
+        (stream->next_frame - stream->this_frame) < 48) {
+        return -1.0; /* Too small buffer to get any necessary headers */
+    }
+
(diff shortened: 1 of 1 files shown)


slouken:
Thanks!

sezero:
@slouken (and @Wohlstand): How well did you test this?  See madplay tag.c:tag_parse() for possible gotchas.

slouken:
I didn't test it, I relied on this being mature code in the MixerX fork.

Do you have test cases where this doesn't work or is worse than not having it implemented at all?

sezero:
> I didn't test it, I relied on this being mature code in the MixerX fork.

IMO, not a good idea.

> Do you have test cases where this doesn't work or is worse than not having it implemented at all?

I don't. I haven't got the time to thoroughly reviewing / testing it yet.

Wohlstand:
> @slouken (and @Wohlstand): How well did you test this? See madplay tag.c:tag_parse() for possible gotchas.

`// Checked out the madplay's source`
Wow, why you didn't mention it to me a year ago? I see it also does parse of the LAME tag also.

> test cases where this doesn't work or is worse than not having it implemented at all?

Speaking about Duration, it may improperly compute the duration of the file contains an invalid Xing/IMFO/VBRI tag, doesn't contain it, contains a false value, etc. Also, it fails to compute the duration when playing Frankenstein streams (a concoction of two or more different-type files). Position reporting will work properly as it simply counts the played amount of PCM samples. In the past, I had a different duration code that makes the full scan of the song and computes the actual duration. This way had a disadvantage: the file got being loaded longer due to this scan. For the VBR streams, if the VBRI tag is broken, the computed duration may be wrong without the full scan of the song.

sezero:
> `// Checked out the madplay's source`
> Wow, why you didn't mention it to me a year ago?

I remember that I did, but whatever.

Wohlstand:
> I remember that I did, but whatever.

I don't remind when and where... I did the check of my mailbox archive, and I see nothing on that...

The only thing you gave me, 4 links to various discussions and specifications against Xing support, etc.

sezero:
> The only thing you gave me, 4 links to various discussions and specifications against Xing support, etc.

I said `whatever`, didn't I..

In any case, madplay source is GPL (_not_ LGPL), so be careful when using stuff from it.

Wohlstand:
> In any case, madplay source is GPL (not LGPL), so be careful when using stuff from it.

if I use some, I would just look at the methods, but code the new with no copypasting, the same as a retelling the plot of the book. At least, even this I sent isn't ideal, it just works for ~1.5 years (at my side I do use the libmad by default until I'll develop the proper CMake build for the mpg123, people didn't blame the bad workflow of MP3 playback at least).

Wohlstand:
Anyway, a bit later I'll try to add more checks and the support for the LAME header in the next patch...

sezero:
One mp3 I have reports 56.607347s duration instead of about 31.320816s
with libmad backend: I can attach the mp3 in question here.

slouken:
> One mp3 I have reports 56.607347s duration instead of about 31.320816s with libmad backend: I can attach the mp3 in question here.

Can you enter a new issue for this, for tracking purposes?

sezero:
Done: https://github.com/libsdl-org/SDL_mixer/issues/403

PR #301, opened 2021-02-21
CMakeLists.txt: expose include directory as target_include_directories

kichikuou:
So that users of this target can write `#include <SDL_mixer.h>` instead of `#include <include/SDL_mixer.h>`.


sezero:
Thanks

PR #303, opened 2021-03-17
native_midi_macosx.c: fixed the C90 warning

Wohlstand:
Isn't should be the debug log print here? Right now, this makes some errors when using C90 compatibility being enforced.


sezero:
@slouken: That is a scary fixme. What to do about it?

Wohlstand:
I opened the #305 instead of this. Mainly to don't trash the "master" branch on my end while synchronizing and keeping every pull-request being clear and contain only related changes implemented at the topic-branch.

PR #304, opened 2021-03-21
Timidity: Fixed the music playing stuck after Timidity song stop

Wohlstand:
When you do stop the song played by Timidity, it will still return the "is Playing = 1", which will lock the ability to start any other music playing at all. To resolve the problem I implemented the `Timidity_Stop()` call that sets the flag into zero that fixes the issue.


sezero:
Your patch seems incomplete. Pushed 620ebbbb01fb9cc7864c9f4acf710c6c543a37f9 as the fix

Wohlstand:
Damn, forgot to backport the implementation of `TIMIDITY_Stop()` at Timidity codec... But glad that you finished the thing, thanks! :fox_face: :+1:

PR #307, opened 2021-04-03
music_fluidsynth.c: Refactor the FLUIDSYNTH_LoadMusic() call

Wohlstand:
The original code of the FLUIDSYNTH_LoadMusic() looks like the scaring lasagna with the bad visuality which is a source of various memory leak errors caused by the human factor.

I made it look better and move all clean-up into the `FLUIDSYNTH_Delete()` by adding all checks for every stage in the middle.

--- 
P.S. In addition I want to allow the FluidSynth to work with the Float32 format too (it can!) (when the Mixer was open with the AUDIO_F32-like format). This should avoid clipping when using the AUDIO_S16 as a source format. I'll make this change as a separate pull request which will depend on this.

### DIFF
--- a/src/codecs/music_fluidsynth.c
+++ b/src/codecs/music_fluidsynth.c
@@ -128,6 +128,7 @@ static void FLUIDSYNTH_Unload()
 
 typedef struct {
     fluid_synth_t *synth;
+    fluid_settings_t *settings;
     fluid_player_t *player;
     SDL_AudioStream *stream;
     void *buffer;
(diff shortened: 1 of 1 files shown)


sezero:
Looks fair.

I suggest renaming 'song_buffer' to 'rw_mem', removing a useless (void*)
cast, and using goto statements instead of repeating the cleanup calls.

A patch against your version is attached.

[patch.txt](https://github.com/libsdl-org/SDL_mixer/files/6253119/patch.txt)

Wohlstand:
Oooke, wait a sec I'll bring up these changes to here...

Wohlstand:
@sezero, done!

Wohlstand:
Whoops, the small mistake I fixed in the condition: it should be `!ret` rather `ret < 0` because I accidentally copy-pasted it from my FluidLite that uses an external MIDI sequencer implementation, it returns -1 when fails to load the song.

sezero:
Found a bug where fluid_player_add_mem() result is checked.
A patch against your latest version is attached (which also removes
a few unnecessary blank lines.)
[patch2.txt](https://github.com/libsdl-org/SDL_mixer/files/6253145/patch2.txt)

sezero:
I guess you beat me to it. Nevertheless, I'll apply yours and then adjust style in a later commit.

Wohlstand:
> Found a bug where fluid_player_add_mem() result is checked.

Yeah, that `ret < 0` was is the same. Gonna to apply your changes.

Okay, gonna prepare the F32 input support...

Wohlstand:
Thanks for the merge! :fox_face: :+1:

sezero:
Friendly request: SDL-1.2 branch needs backporting stuff starting from commit bec0250e9b33c
(but it's different, e.g. it uses SDL_AudioCVT, and other differences.) If you ever have time, can
you backport to SDL-1.2 branch? (Definitely not high priority, and it's OK if you say no.)

Wohlstand:
Would get a shot, that shouldn't be hard :fox_face:

PR #308, opened 2021-04-03
music_fluidsynth.c: Allow the F32 source format support

Wohlstand:
FluidSynth is able to return the Float32 format by itself. When Mixer is open as F32, it's better to stream the output as F32 too:
- no need to convert the sample format when F32 format is used as output
- no clipping caused by using of the S16 as the source format

### DIFF
--- a/src/codecs/music_fluidsynth.c
+++ b/src/codecs/music_fluidsynth.c
@@ -56,6 +56,7 @@ typedef struct {
     void (*fluid_synth_set_gain)(fluid_synth_t*, float);
     int (*fluid_synth_sfload)(fluid_synth_t*, const char*, int);
     int (*fluid_synth_write_s16)(fluid_synth_t*, int, void*, int, int, void*, int, int);
+    int (*fluid_synth_write_float)(fluid_synth_t*, int, void*, int, int, void*, int, int);
     fluid_player_t* (*new_fluid_player)(fluid_synth_t*);
     fluid_settings_t* (*new_fluid_settings)(void);
     fluid_synth_t* (*new_fluid_synth)(fluid_settings_t*);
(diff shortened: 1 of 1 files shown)


sezero:
This looks unnecessarily complex. Will try revising and run-testing.

sezero:
It should be as simple as the attached patch - unless I am missing something.
(patch is against unmodified mainstream SDL_mixer.)

[patch2.txt](https://github.com/libsdl-org/SDL_mixer/files/6253213/patch2.txt)

Wohlstand:
I guess, yeah, no need to define all non-32 types to just find few 32-types, you right here :fox_face:

sezero:
And the write functions you suggested were trying things really hard:
the _s16 and _float signatures are the same, so we should be good.

Can you test and give me the OK, so that I can apply?

Wohlstand:
> Can you test and give me the OK, so that I can apply?

I am in process of preparing for that... Once I'll test, I'll put your update to my side, so you will be able to review and accept.

sezero:
Minorly revised patch attached.

[patch3.txt](https://github.com/libsdl-org/SDL_mixer/files/6253229/patch3.txt)

Wohlstand:
> const Uint8 channels = 2;

Shouldn't it fail on some old compilers that don't support the `const` type? But ok, that fine, I totally forgot that there at code a lot of `const` usages here. That bad think has prevented me to add the `const` before.

Wohlstand:
However, this one patch I tested, it seems work. Gonna apply your lates fixes...

Wohlstand:
Done!

Wohlstand:
WAIT!!!! I found the error: You still initialize the buffer with the size of a Sint64, you should initialize the same buffer with the size of a `float` when the source sample format is F32!!! That will cause the SIGSEGV in some cases. I wonder why it didn't happen to me...

Wohlstand:
Okay, I miss-looked... That you writing the size, and then you do 2x increase it when the F32 is used, and then you do initialization. All fine.

sezero:
Applied my patch with further revision as https://github.com/libsdl-org/SDL_mixer/commit/f6cdc3667fbce7467b1ae23cc67c1852c4a482b9

PR #309, opened 2021-04-03
.gitignore instead of .hgignore

Wohlstand:
As the little begin


Wohlstand:
This update has gone into #310

PR #313, opened 2021-04-10
Add 'Libs.private' field to pkg-config file

ffontaine:
In order to support static linking, SDL_mixer.pc should include a
'Libs.private' field listing all the libraries that SDL_mixer requires.

This patch adds such a field and also modifies configure.ac so that
EXTRA_LDFLAGS (which is now also used as the value of 'Libs.private')
no longer includes SDL_LIBS. This is done so as to prevent libraries
required by SDL from being listed twice when 'pkg-config --libs --static
SDL_mixer' is run (they're already shown because of the 'Requires: sdl'
line in SDL_mixer.pc). Makefile.in is also adjusted accordingly.

Upstream status: submitted
https://bugzilla.libsdl.org/show_bug.cgi?id=3278

Signed-off-by: Rodrigo Rebello <rprebello@gmail.com>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/sdl_mixer/0001-Add-Libs.private-field-to-pkg-config-file.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -61,7 +61,7 @@ $(objects)/.created:
 .PHONY: all install install-hdrs install-lib install-bin uninstall uninstall-hdrs uninstall-lib uninstall-bin clean distclean dist
 
 $(objects)/$(TARGET): $(OBJECTS) $(VERSION_OBJECTS)
-	$(LIBTOOL) --mode=link $(CC) -o $@ $(OBJECTS) $(VERSION_OBJECTS) $(LDFLAGS) $(EXTRA_LDFLAGS) $(LT_LDFLAGS)
+	$(LIBTOOL) --mode=link $(CC) -o $@ $(OBJECTS) $(VERSION_OBJECTS) $(LDFLAGS) $(SDL_LIBS) $(EXTRA_LDFLAGS) $(LT_LDFLAGS)
 
 $(objects)/playwave$(EXE): $(objects)/playwave.lo $(objects)/$(TARGET)
 	$(LIBTOOL) --mode=link $(CC) -o $@ $(objects)/playwave.lo $(SDL_CFLAGS) $(objects)/$(TARGET) $(SDL_LIBS) $(LDFLAGS)
--- a/SDL2_mixer.pc.in
+++ b/SDL2_mixer.pc.in
@@ -8,5 +8,6 @@ Description: mixer library for Simple DirectMedia Layer
 Version: @VERSION@
 Requires: sdl2 >= @SDL_VERSION@
 Libs: -L${libdir} -lSDL2_mixer
+Libs.private: @EXTRA_LDFLAGS@
 Cflags: -I${includedir}/SDL2
 
--- a/configure.ac
+++ b/configure.ac
@@ -194,7 +194,6 @@ AM_PATH_SDL2($SDL_VERSION,
             AC_MSG_ERROR([*** SDL version $SDL_VERSION not found!])
 )
 EXTRA_CFLAGS="$EXTRA_CFLAGS $SDL_CFLAGS"
-EXTRA_LDFLAGS="$EXTRA_LDFLAGS $SDL_LIBS"
 
 dnl Check for math library
 AC_CHECK_LIB(m, pow, [LIBM="-lm"])


Radvendii:
I just stumbled across this myself, and the solution I came up with involved `Requires.private` as well, for dependencies that have pkgconfig files (`vorbisfile`, `vorbisenc`, `vorbis`, `opusfile`, `opus`). `-lFLAC` still needed to be in `Libs.private`.

Should we be doing that instead? I was also hard-coding things, so I don't know how easy it would be to modify this solution to use `Requires.private`

ffontaine:
Sure, using `Requires.private` instead of `Libs.private` is also perfectly fine but this will have more impacts on `configure.ac` as we'll need to add a dedicated field (e.g. `SDL_REQUIRES_PRIVATES`) to save the list of pkg-config dependencies names.

Radvendii:
Oh, okay; that's fair.

I just tried your patch and am still unable to compile something statically against SDL2_mixer. Inspecting the `.pc` file, I think it's bceause `Libs.private` does not include transitive dependencies. For instance, it has `-lopusfile`, but not `-lopus`, on which it depends. It also doesn't have `-logg`.

Have you been able to get things statically linking against SDL2_mixer using this patch? I wonder if something is messed up on my setup.

ffontaine:
We have this patch for more than 5 years in buildroot but only for **sdl_mixer 1.2** (i.e. commit d1725fcb7c4e987aeb7ecdc94cb8b6375b702170). We have no build failures with sdl_mixer on the autobuilders.

Static building of sdl2_mixer is disabled because it depends on sdl2 on which static building is disabled since its addition 5 years ago (I don't know why).

Concerning, your question, `Libs.Private` will include transitive dependencies if they're correctly retrieved through pkg-config. For example, `-logg` should be in `OPUSFILE_LIBS` thanks to this call
https://github.com/SDL-mirror/SDL_mixer/blob/2d53d383a63293aff1249d525ff978e3a4dac002/configure.ac#L710

If it is not added, then I would advise to check the content of `opusfile.pc` to be sure that it contains `-logg`. That being said, opusfile is not a dependency of sdl 1.2 branch so it is possible that there is an issue.

Radvendii:
You're right. `opusfile.pc` is missing `-lopus` and `-logg`. Thank you.

ffontaine:
A possible issue is that pkg-config must be called with `--static` to retrieve Private fields. In buildroot, we have a wrapper to handle that. You should check the difference between `pkg-config --libs opusfile `and `pkgconfig --libs --static opusfile`

Radvendii:
Yeah, using `pkg-config --libs --cflags --static opusfile` gives the correct output, so I don't understand why it's not showing up in the SDL_mixer `.pc` file. Although I'm also not seeing `-lFLAC`, which is explicitly added here
https://github.com/libsdl-org/SDL_mixer/blob/master/configure.ac#L622
So I'm not sure what's going on. Still digging. I have almost no experience with autotools, so it's slow going.

ffontaine:
If you have no experience with autotools, may I suggest that you forgot to run `autoreconf`?

sezero:
>  forgot to run `autoreconf`?

In SDL_mixer, `autoreconf` is not supported: one should use the included `autogen.sh` script.

Radvendii:
> If you have no experience with autotools, may I suggest that you forgot to run `autoreconf`?

I was not running that, thanks for the tip.

> In SDL_mixer, `autoreconf` is not supported: one should use the included `autogen.sh` script.

And thanks for cutting short a rabbit hole of `autoreconf` errors :stuck_out_tongue: 

Still no `-lopusfile`, `-logg` or `-lFLAC` though. Which makes sense, since the only update to `configure.in` was *removing* stuff from `EXTRA_LDFLAGS`.

ffontaine:
Did you set `--disable-music-flac-shared`? By default, it is set to true, so if your system embeds a static and a shared library, sdl_mixer will prefer the shared one. This is the same thing for `--disable-music-opus-shared`

Radvendii:
Yes, `--disable-music-flac-shared` and `--disable-music-opus-shared` are set.

Radvendii:
Is this verified to output the right Libs.private for you? Because if so, it's probably an issue with my system and shouldn't stop this PR going through.

Radvendii:
By contrast the `mpg123` dependency disables support when the symbol is missing:
https://github.com/libsdl-org/SDL_mixer/blob/master/configure.ac#L669-L696

I'm not sure if one of these is a mistake / oversight or there's some difference between the libraries that's the reason for it.

PR #331, opened 2021-09-19
Add source files to macOS static library build

lufinkey:
The source files were missing from the macos xcode project for the static library target, so I added them


sezero:
Can someone (@slouken, @icculus) please verify this?

icculus:
Tried this on Xcode here, and it builds a static library, so I guess we're good.   :)

PR #337, opened 2021-11-22
Added support of quoted strings in Timidity config

### LINKED ISSUE #334: Add support of quoted strings in Timidity config
Timidity config parameters `dir` and `source` accept a set of paths separated by spaces as arguments. But if paths themselves contain spaces they become broken.
The obvious fix is to just write paths as quoted strings however it doesn't work. The quotation mark character isn't processed by SDL Mixer's Timidity.

Dasperal:
Resolves #334

An argument of property in Timidity config can be surrounded with `"` or `'` characters to prevent splitting into several arguments if it contains spaces or other delimiter characters.

If an argument itself contains `"` characters it should be surrounded with `'` characters and vice versa.

### DIFF
--- a/src/codecs/timidity/timidity.c
+++ b/src/codecs/timidity/timidity.c
@@ -97,10 +97,62 @@ static int read_config_file(const char *name, int rcf_count)
     if (*w[0] == '#')
         continue;
 
-    while (w[words] && *w[words] != '#') {
-      if (++words == MAXWORDS) break;
-      w[words]=SDL_strtokr(NULL, " \t\240", &endp);
+    while (words < MAXWORDS - 1) /* -1 : next arg */
+    {
+      while (*endp == ' ' || *endp == '\t' || *endp == '\240')
+        endp++;
+      if (*endp == '\0' || *endp == '#')
+        break;
+      if (*endp == '"' || *endp == '\'') /* quoted string */
+      {
+        char *terminator;
+        if ((terminator = SDL_strchr(endp + 1, *endp)) != NULL) /* terminated */
+        {
+          if (terminator[1] == ' ' || terminator[1] == '\t' || terminator[1] == '\240' || terminator[1] == '\0')
+          {
+            char *extraQuote;
+            if ((extraQuote = SDL_strchr(endp + 1, *endp == '"' ? '\'' : '"')) != NULL && extraQuote < terminator)
+            {
+                SNDDBG(("%s: line %d: Quote characters are not allowed inside a quoted string", name, line));
+                goto fail;
+            }
+
+            w[++words] = endp + 1;
+            endp = terminator + 1;
+            *terminator = '\0';
+          }
+          else /* no space after quoted string */
+          {
+            SNDDBG(("%s: line %d: There must be at least one whitespace between string terminator (%c) and the next parameter", name, line, *endp));
+            goto fail;
+          }
+        }
+        else /* not terminated */
+        {
+          SNDDBG(("%s: line %d: The quoted string is not terminated", name, line));
+          goto fail;
+        }
+      }
+      else /* not quoted string */
+      {
+        w[++words] = endp;
+        while (!(*endp == ' ' || *endp == '\t' || *endp == '\240' || *endp == '\0'))
+        {
+          if (*endp == '"' || *endp == '\'') /* no space before quoted string */
+          {
+              SNDDBG(("%s: line %d: There must be at least one whitespace between previous parameter and a beginning of the quoted string (%c)", name, line, *endp));
+              goto fail;
+          }
+          endp++;
+        }
+        if (*endp != '\0') /* unless at the end-of-string (i.e. EOF) */
+        {
+          *endp = '\0';    /* terminate the token */
+          endp++;
+        }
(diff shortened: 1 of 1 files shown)


sezero:
Does this match Timidity++?

Dasperal:
It is based on Timidity++ but matches the original SDL_Mixer's Timidity behavior except for quoted strings. And it appears that space between a quoted string and the next argument became optional.

Do I need to make config file parsing fully Timidity++ compatible?

sezero:
> It is based on Timidity++ but matches the original SDL_Mixer's Timidity behavior except for quoted strings.

OK, 

> And it appears that space between a quoted string and the next argument became optional.

Eh? Say, `0 "instruments/acpiano"amp=125` is valid now?

> Do I need to make config file parsing fully Timidity++ compatible?

No, only and only quoted string _so that_ paths with spaces would be handled correctly.
And please test carefully - e.g. if we hit a `"` do we require it closing match, etc?

Dasperal:
Updated patch. Fixed bug which stops processing of the line after an unquoted argument. Didn't notice it originally because I had a Timidity config with properties with only one argument.

Tested following scenarios of splitting line by words:
1) line `dir abc def ghi jkl`
will be split to `dir`, `abc`, `def`, `ghi`, `jkl`
2) line `dir "abc def" ghi jkl`
will be split to `dir`, `abc def`, `ghi`, `jkl`
3) line `dir "abc"def "ghi"jkl`
will be split to `dir`, `abc`, `def`, `ghi`, `jkl`
4) line `dir "abc""def""ghi""jkl"`
will be split to `dir`, `abc`, `def`, `ghi`, `jkl`
5) line `dir abc"def" ghi"jkl"`
will be split to `dir`, `abc"def"`, `ghi"jkl"`
6) line `dir "abc def ghi jkl`
will be split to `dir`, `"abc`, `def`, `ghi`, `jkl`
7) line `dir "abc def ghi' jkl`
will be split to `dir`, `"abc`, `def`, `ghi'`, `jkl`
8) line `dir 'abc "def" ghi' jkl`
will be split to `dir`, `abc "def" ghi`, `jkl`

I think scenarios 3, 4, 6, 7 should be errors.

sezero:
> I think scenarios 3, 4, 6, 7 should be errors.

Indeed.

Case 8 is weird, though. Anyone does that?

sezero:
My point is, only the first scenario is sensible (unless I'm proven otherwise),
e.g. path arguments to `dir`.  So we should do something like, in pseudo-code:
```c
  seenquote  = 0;
  [hit 1st quote]
  if (!isspace(*beginquote)) {
    error; /* only arguments can be quoted */
  }
  seenquote++;
  [find endquote]
  if (!endquote) {
    error;
  }
  [then :]
  if (seenquote > 2) {
    error; /* multiple quoted args: shouldn't be possible, yes? */
  }
```

Dasperal:
>Why do you want to accept case 5??

In the original implementation with "SDL_strtokr", the resulting words may contain any garbage characters. In this particular case, it may look like "You wanted a quoted argument but forgot a space" so I can make this an error.

>Case 8 is weird, though. Anyone does that?

I don't know why it can be needed nor if it ever will.

Errors are just
``` C
SNDDBG(("%s: line %d: Error message", name, line));
goto fail;
```
right?

sezero:
> > Why do you want to accept case 5??
> 
> In the original implementation with "SDL_strtokr", the resulting words may contain any garbage characters. In this particular case, it may look like "You wanted a quoted argument but forgot a space" so I can make this an error.

Let's make that one an error

> > Case 8 is weird, though. Anyone does that?
> 
> I don't know why it can be needed nor if it ever will.

Indeed, let's make that an error too

> Errors are just
> ```c
> SNDDBG(("%s: line %d: Error message", name, line));
> goto fail;
> ```
> right?

Yes.

Dasperal:
No response to the last push so far, so I think I should leave a comment.

Updated patch. Mentioned cases (3 - 8) are treated as errors now.

Is there anything else I need to do or test?

sezero:
Sorry, was a little busy + lazy + etc

Will test soon and if no issues, will merge.  Would you agree that I apply this to [libtimidity](https://github.com/sezero/libtimidity) and [SDL_sound](https://github.com/icculus/SDL_sound) too?

Dasperal:
> Would you agree that I apply this to [libtimidity](https://github.com/sezero/libtimidity) and [SDL_sound](https://github.com/icculus/SDL_sound) too?

Yes

sezero:
@icculus, @slouken: Does this patchset look good to you? Do you see any gotchas?

slouken:
Yep, looks good, thanks!

sezero:
Patch now applied to SDL_sound.

sezero:
Patch now applied to libtimidity too.

sezero:
@Dasperal, @slouken: I pushed https://github.com/libsdl-org/SDL_mixer/commit/d1cff77caaf84783b510d603e466b566852a4e94 for better readability: please make sure that I didn't screw-up.

Dasperal:
Tested. Quoted strings still work

PR #343, opened 2021-12-12
configure.ac cleanup

ThomasKuehne:
in configure.ac: 
- add missing AC_PROG for gcc-fat.sh
- add missing AC_PROG for ltmain.sh
- replace obsolete AC_TRY_COMPILE
- replace obsolete AC_TRY_RUN
- replace obsolete AC_HELP_STRING


sezero:
- `AC_PROG_CPP` and `AC_PROG_RANLIB`B well, maybe, but we really don't need `AC_PROG_CXX`
- I do not get the warnings you are mentioning in your commit messages?

sezero:
I just replaced the deprecated `AC_` macros in git master.

All that remains is the last two commits in your patchset: I need real
convincing, as I mentioned above.

ThomasKuehne:
those issues where identified by running
- autoscan-2.71
- autoupdate-2.71 --force

sezero:
They really aren't needed, and at the very least `AC_PROG_CXX` is a false alarm from autoscan.
Closing this.

And thanks for reminding about the obsolete autoconf macros which are now replaced.

PR #344, opened 2021-12-12
resolve -Wmissing-field-initializers warnings

ThomasKuehne:
fixes:
src/codecs/music_mpg123.c:66:1: warning: missing field 'mpg123_close' initializer [-Wmissing-field-initializers]
src/codecs/music_ogg.c:66:1: warning: missing field 'ov_clear' initializer [-Wmissing-field-initializers]

### DIFF
--- a/src/codecs/music_mpg123.c
+++ b/src/codecs/music_mpg123.c
@@ -62,7 +62,11 @@ typedef struct {
 } mpg123_loader;
 
 static mpg123_loader mpg123 = {
-    0, NULL
+    0, NULL,
+    NULL, NULL, NULL, NULL, NULL,
+    NULL, NULL, NULL, NULL, NULL,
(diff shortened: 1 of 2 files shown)


sezero:
I guess you are running with `-pedantic`: Don't, it is overkill.
We don't need this one.

slouken:
Missing field initializers are undefined. In this case since the object is static, the entire object is initialized to zero, so we probably don't need any initializers, but if we have some, we should have them all. If this object were on the stack, this would be a trap waiting to spring on someone in the future.

sezero:
>  the object is static, the entire object is initialized to zero, so we probably don't need any initializers,

Exactly: What we had here is very common practice. If one is really
worried, then the structure can be zeroed during init. I really don't see
why we need this patch.

PR #345, opened 2021-12-12
fix memory leaks in Parse* (music_wav.c)

pionere:


sezero:
@Wohlstand: This looks correct to me. Any comments?

Wohlstand:
Looks fine, really, there were memory leaks that will happen on a read failure.

sezero:
Patch applied, thanks.

PR #347, opened 2021-12-13
minor cleanup of init_position_args

pionere:
- no need to zero fill the position_args struct, since all fields are set in the subsequent lines
- no need to cast &args->channels

### DIFF
--- a/src/effect_position.c
+++ b/src/effect_position.c
@@ -1548,14 +1548,13 @@ static void SDLCALL _Eff_position_f32sys_c6(int chan, void *stream, int len, voi
 
 static void init_position_args(position_args *args)
 {
-    SDL_memset(args, '\0', sizeof (position_args));
     args->in_use = 0;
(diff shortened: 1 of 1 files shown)


sezero:
> * no need to zero fill the position_args struct, since all fields are set in the subsequent lines

Or maybe we could have kept the memset but remove the two 0 assignments.
In the very unlikely chance that the structure changes, the memset would be still
good though

>     * no need to cast &args->channels

```
src/effect_position.c: In function 'init_position_args':
src/effect_position.c:1557: warning: passing argument 3 of 'Mix_QuerySpec' discards qualifiers from pointer target type
include/SDL_mixer.h:161: note: expected 'int *' but argument is of type 'volatile int *'
```
(Maybe I should add the gc warning options to configure.ac -- it
doesn't seem to add `-Wall` by default.)

Let's not apply this patch for now.

PR #348, opened 2021-12-13
fix memory leak in WAV_CreateFromRW

pionere:


sezero:
@Wohlstand: You touched the wav source -- any gotchas here?

Wohlstand:
Here is the only possible memory leak because of unclear loops and meta-tags that was left after failure (when loops and/or tags was been parsed before the error has occurred)

Wohlstand:
Anyway, these things should be analyzed using static analyzers, otherwise, you'll get the effect called "глаз замылился" that means, one person can recheck the same thing with hundreds of attempts and find no errors, but once another person will check, the bug will be found. Also, some cases were hard being recognized by humans, and therefore static analyzers would help to figure out these hard cases.

PR #349, opened 2021-12-14
do not report a non-playing channel as paused

pionere:


sezero:
@slouken and/or @icculus: this is for you.

PR #350, opened 2021-12-14
Some CMake tweaks

Daft-Freak:
Some small CMake changes from an old branch. Also fixes projects that use `add_subdirectory` for SDL+mixer (can't use `find_package`, no `_INCLUDE_DIR/_LIBRARY`)... and then re-fixes the action because it has an old CMake config without the targets...


sezero:
Thanks

PR #352, opened 2021-12-15
improve ParseFMT

pionere:
- ensure the structs used in ParseFMT are packed without padding
- use stack to load the format information
- ensure ParseFMT does not read outside of the allocated memory in case of EXT_CODE

### DIFF
--- a/src/codecs/music_wav.c
+++ b/src/codecs/music_wav.c
@@ -79,6 +79,7 @@ typedef struct {
 #define WAVE_MONO   1
 #define WAVE_STEREO 2
 
+#pragma pack(push, 1)
 typedef struct {
 /* Not saved in the chunk we read:
     Uint32  chunkID;
@@ -90,7 +91,7 @@ typedef struct {
     Uint32  byterate;       /* Average bytes per second */
     Uint16  blockalign;     /* Bytes per sample block */
     Uint16  bitspersample;      /* One of 8, 12, 16, or 4 for ADPCM */
-} WaveFMT;
+} WaveFMTHeader;
 
 typedef struct {
     Uint16  cbSize;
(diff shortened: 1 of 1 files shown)


src/codecs/music_wav.c
```diff
@@ -90,7 +91,7 @@ typedef struct {
     Uint32  byterate;       /* Average bytes per second */
     Uint16  blockalign;     /* Bytes per sample block */
     Uint16  bitspersample;      /* One of 8, 12, 16, or 4 for ADPCM */
-} WaveFMT;
+} WaveFMTHeader;
 
 typedef struct {
     Uint16  cbSize;
```
slouken, line 97: Suggest calling this structure WaveFMTEx, moving cbSize to WaveFMT above, and including WaveFMT here as 'format'.
This more closely reflects Microsoft's definitions:
https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex

pionere: I have a few examples of WAV where the size of the FMT chunk is only 16. If the cbSize is moved above the logic has to test for a 'magic' number instead of the size of the struct.

sezero: I think @Wohlstand should join the discussion too.

slouken: > I have a few examples of WAV where the size of the FMT chunk is only 16. If the cbSize is moved above the logic has to test for a 'magic' number instead of the size of the struct.

Fair enough, but let's make the rest of the suggested changes.

pionere: you mean renaming it to WaveFMTEx? what would be the name of the other struct? At the moment it is called WaveFMTex.

Wohlstand: @sezero , @pionere , just now I made some check and I found the "comparison of integers of different signs" warning:

```c
if (bits < sizeof(fmt)) {
```

So, I guess, you should do the next:
```c
if (bits < (int)sizeof(fmt)) {
```
because `sizeof()` always returns the `size_t` type which is usually `unsigned long` on 64bit systems or `unsigned int` on 32-bit.

slouken: Fixed, thanks!

PR #354, opened 2021-12-15
mixer.c: Fixed the crash at the Mix_LoadMusic_RW

Wohlstand:
I found this crash case a while ago at MixerX. It happen because the "fragment" wasn't properly updated and left being NULL. That leads an attempt to dereferrence it while it has NULL inside.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -636,7 +636,7 @@ static SDL_AudioSpec *Mix_LoadMusic_RW(SDL_RWops *src, int freesrc, SDL_AudioSpe
     Mix_UnlockAudio();
 
     if (count > 0) {
-        *audio_len = (count - 1) * fragment_size + fragment->size;
+        *audio_len = (count - 1) * fragment_size + last->size;
         *audio_buf = (Uint8 *)SDL_malloc(*audio_len);
         if (*audio_buf) {
             Uint8 *dst = *audio_buf;


src/mixer.c
```diff
@@ -597,12 +597,14 @@ static SDL_AudioSpec *Mix_LoadMusic_RW(SDL_RWops *src, int freesrc, SDL_AudioSpe
...
         if (!fragment) {
             /* Uh oh, out of memory, let's return what we have */
+            fragment = last;
             break;
         }
         fragment->data = (Uint8 *)SDL_malloc(fragment_size);
         if (!fragment->data) {
             /* Uh oh, out of memory, let's return what we have */
             SDL_free(fragment);
+            fragment = last;
```
sezero, line 607: Haven't followed the code carefully, but: there is not a chance that
`last` and `fragment` would point to the same memory here is there?
If that ever happens, it would lead to a double-free.

sezero:
@slouken ?

src/mixer.c
Wohlstand: I guess it's better to replace the "fragment" with "last" at the final check?

```c
    if (count > 0 && last) {
        *audio_len = (count - 1) * fragment_size + last->size; /* <------- THIS ONE*/
        *audio_buf = (Uint8 *)SDL_malloc(*audio_len);
        if (*audio_buf) {
            Uint8 *dst = *audio_buf;
            for (fragment = first; fragment; fragment = fragment->next) {
                SDL_memcpy(dst, fragment->data, fragment->size);
                dst += fragment->size;
            }
        } else {
            SDL_OutOfMemory();
            spec = NULL;
        }
    } else {
        Mix_SetError("No audio data");
        spec = NULL;
    }
```

Wohlstand: Mainly, because this code block adds the size of the last fragment to finalize the data block, using the last fragment. But, because of the out-of-memory failure from above, the fragment is expected to be something existing because of the `fragment->size`. So, I guess, that's really reasonable to use the `last->size` instead of the `fragment->size` here and remove `fragment = last` lines from above.

Wohlstand:
Okay, I simplified the solution

src/mixer.c
```diff
@@ -635,8 +635,8 @@ static SDL_AudioSpec *Mix_LoadMusic_RW(SDL_RWops *src, int freesrc, SDL_AudioSpe
 
     Mix_UnlockAudio();
 
-    if (count > 0) {
-        *audio_len = (count - 1) * fragment_size + fragment->size;
+    if (count > 0 && last) {
```
slouken, line 638: Aren't we guaranteed that last will not be NULL if count > 0?

Wohlstand: > Aren't we guaranteed that last will not be NULL if count > 0?

I guess, yes, you're right, then, the "last" check can be easily removed.

Wohlstand: Made

slouken: Perfect! :)

PR #355, opened 2021-12-15
Remove all initializers from static loader arrays

Wohlstand:
Addition to https://github.com/libsdl-org/SDL_mixer/commit/143b0bbfd50654af667c87143c1a2d69ef80f151

EDIT: As was discussed below, I changed the entire goal of this PR:
----
C standard gives guarantee they will be NULLed by default:
https://port70.net/~nsz/c/c89/c89-draft.html#3.5.7

> If an object that has static storage duration is not initialized explicitly, it is initialized implicitly as if every member that has arithmetic type were assigned 0 and every member that has pointer type were assigned a null pointer constant. If an object that has automatic storage duration is not initialized explicitly, its value is indeterminate.

### DIFF
--- a/src/codecs/music_flac.c
+++ b/src/codecs/music_flac.c
@@ -69,9 +69,7 @@ typedef struct {
                         FLAC__MetadataType type);
 } flac_loader;
 
-static flac_loader flac = {
-    0, NULL
-};
+static flac_loader flac;
 
 #ifdef FLAC_DYNAMIC
 #define FUNCTION_LOADER(FUNC, SIG) \
--- a/src/codecs/music_fluidsynth.c
+++ b/src/codecs/music_fluidsynth.c
@@ -62,9 +62,7 @@ typedef struct {
     fluid_synth_t* (*new_fluid_synth)(fluid_settings_t*);
 } fluidsynth_loader;
 
-static fluidsynth_loader fluidsynth = {
-    0, NULL
-};
+static fluidsynth_loader fluidsynth;
 
 #ifdef FLUIDSYNTH_DYNAMIC
 #define FUNCTION_LOADER(FUNC, SIG) \
--- a/src/codecs/music_mikmod.c
+++ b/src/codecs/music_mikmod.c
@@ -69,9 +69,7 @@ typedef struct {
     UBYTE* md_volume;
 } mikmod_loader;
 
-static mikmod_loader mikmod = {
-    0, NULL
-};
+static mikmod_loader mikmod;
 
 #ifdef MIKMOD_DYNAMIC
 #define FUNCTION_LOADER(FUNC, SIG) \
--- a/src/codecs/music_modplug.c
+++ b/src/codecs/music_modplug.c
@@ -47,10 +47,7 @@ typedef struct {
     const char* (*ModPlug_GetName)(ModPlugFile* file);
 } modplug_loader;
 
-static modplug_loader modplug = {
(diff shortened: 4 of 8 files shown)


slouken:
Actually let's just remove all the static initializers. That way if we change the decoder structures they'll still be zeroed with no additional work.

Wohlstand:
> Actually let's just remove all the static initializers. That way if we change the decoder structures they'll still be zeroed with no additional work.

Do you mean I should remove these initializers and just add the `SDL_memset(.., 0, sizeof(..)` at beginning of every initializer?

Also, that needs also make the "loaded" field being off the structure and being statically initialized as 0, otherwise, if structure is not initialized, then, garbage will make the illusion like library was been initialized, so, make the sort of this:

изображение

---
The fact: I reminded some other but similar case: Before you did major rework at mainstream Mixer, I had developed my own variant of improvement and I had a different way, so, lemme find and show you some (maybe revive that my concept to here?)

It's about the `Mix_MusicInterface` initializer, and I had it being dynamically initialized in the next way:

- every module has its own "init interface" call
- it makes SDL_memset(interface, 0, sizeof(interface)) that fills all fields as null
- only used and only necessary calls were shown here
- once somebody adds a new field into Mix_MusicInterface, no need to change ~~thousands~~ all other modules to add "NULL"s into it, that's generally a pain.

изображение

slouken:
No, I mean:
```
static fluidsynth_loader fluidsynth;
```
All the fields are initialized to 0 and NULL by C language definition. We don't need to provide initializers unless we want non-zero values.

Wohlstand:
> All the fields are initialized to 0 and NULL by C language definition. We don't need to provide initializers unless we want non-zero values.

Very depends on the compiler, because I literally had these structures being filled with garbage. The full zeroing works on debug mode only, on release mode that will just a ton of garbage.

slouken:
They were only filled with garbage because they were partially initialized. If you leave off the initializers entirely, you'll see that they are zeroed.

Wohlstand:
Anyway, I want to make the small check:
- I'll remove initializers
- I'll add SDL_assert_release traps
- I'll build the thing as a release
- Will try to run

Wohlstand:
Anyway, I tried to search around, and people say that if you leave the structure field being just defined, it will be uninitialized at all. Where is exactly described that if you define the static structure field and being untouched, it will be automatically initialized by zeroes?

Wohlstand:
btw, I found something here: https://coderedirect.com/questions/191022/are-the-members-of-a-global-structure-initialized-to-zero-by-default-in-c

The C99 / 2003 was mentioned, but what about C89/C90 ?

Wohlstand:
Okay, I found some:
>  If an object that has static storage duration is not initialized explicitly, it is initialized implicitly as if every member that has arithmetic type were assigned 0 and every member that has pointer type were assigned a null pointer constant. If an object that has automatic storage duration is not initialized explicitly, its value is indeterminate.65

Wohlstand:
@slouken, so, I triple-checked all stuff, and I updated this PR, so, should be fine now :eyes:

slouken:
Looks good, thanks!

sezero:
Oh yes, finally.

@slouken: Merge?

Wohlstand:
> @slouken: Merge?

Already merged :smirk_cat:

sezero:
> Already merged smirk_cat

Damn, missed it :)

PR #356, opened 2021-12-15
music_xmp.c: Fixed the correctness of meta-tags

Wohlstand:
To get the proper title of the song, it should be taken from another place. The field was actually taken is a copyright field.

### DIFF
--- a/src/codecs/music_xmp.c
+++ b/src/codecs/music_xmp.c
@@ -219,8 +219,13 @@ void *XMP_CreateFromRW(SDL_RWops *src, int freesrc)
 
     meta_tags_init(&music->tags);
     libxmp.xmp_get_module_info(music->ctx, &music->mi);
+
+    if (music->mi.mod->name[0]) {
+        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.mod->name);
+    }
+
     if (music->mi.comment) {
-        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.comment);
+        meta_tags_set(&music->tags, MIX_META_COPYRIGHT, music->mi.comment);
     }
 
     if (freesrc) {


sezero:
`xmp_test_module_from_memory` is added to libxmp as of v4.5.0:
this addition brings in a big requirement.

sezero:
Does `music->mi.name` not serve your purpose?

Wohlstand:
> Does music->mi.mod->name not serve your purpose?

Lemme check...

sezero:
I mean something like this:
```diff
diff --git a/src/codecs/music_xmp.c b/src/codecs/music_xmp.c
index b5b321f..babb52a 100644
--- a/src/codecs/music_xmp.c
+++ b/src/codecs/music_xmp.c
@@ -221,8 +221,8 @@ void *XMP_CreateFromRW(SDL_RWops *src, int freesrc)
 
     meta_tags_init(&music->tags);
     libxmp.xmp_get_module_info(music->ctx, &music->mi);
-    if (music->mi.comment) {
-        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.comment);
+    if (music->mi.mod->name[0]) {
+        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.mod->name);
     }
 
     if (freesrc) {
```

Wohlstand:
Nope, it doesn't work, because mi.mid itself is NULL.

sezero:
That's suspicious, not supposed to happen..

Wohlstand:
I meant, it's NULL if I will do:
```c
    if (music->mi.mod && music->mi.mod->name[0]) {
        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.mod->name);
    }
    libxmp.xmp_get_module_info(music->ctx, &music->mi);
```

but will work fine if:
```c
    libxmp.xmp_get_module_info(music->ctx, &music->mi);

    if (music->mi.mod && music->mi.mod->name[0]) {
        meta_tags_set(&music->tags, MIX_META_TITLE, music->mi.mod->name);
    }
```

Wohlstand:
Updated, should be fine

src/codecs/music_xmp.c
```diff
@@ -218,9 +218,15 @@ void *XMP_CreateFromRW(SDL_RWops *src, int freesrc)
     }
 
     meta_tags_init(&music->tags);
+
     libxmp.xmp_get_module_info(music->ctx, &music->mi);
+
+    if (music->mi.mod && music->mi.mod->name[0]) {
```
sezero, line 224: `music->mi.mod` is useless here: if it's null, then we are playing a NULL mod.

src/codecs/music_xmp.c
```diff
@@ -218,9 +218,15 @@ void *XMP_CreateFromRW(SDL_RWops *src, int freesrc)
     }
 
     meta_tags_init(&music->tags);
+
```
sezero, line 221: Additional empty lines not needed, please lose them

Wohlstand: I can remove it, but not lose, I can lose it if I miss-looked while copy-pasting, then I lost the line

Wohlstand: Anyway, I vote that these whitespaces should be because code looks nicer and not so clobbered when no spaces between logical parts, but, especially this line, I guess, it's fine if I remove it

sezero: Couldn't understand a word :)

Anyways, what I meant is don't add new empty lines around there :)

sezero:
OK, applying.

src/codecs/music_xmp.c
Wohlstand: > Couldn't understand a word :)

Because it's rare slang or dialect? I didn't see this use case so much, for it looks just very weird for me...

BTW, as an off-topic (this one is solved as I removed the): We should beautify the entire Timidity as original author nobody cares about it, and we develop an own independent branch of it (for example, using the "Artistic Style" program, configured for the SDL style. I use this tool to quickly beautify code fragments into a specific style, very convenient!) because it looks ugly because of lots reasons:
- Mixed tabs and spaces indention, nobody suggests using this
- Trailing spaces, lots of trailing spaces, they just annoying and being ballast at files

I guess this should become a separate discussion.

sezero: > > Couldn't understand a word :)
> 
> Because it's rare slang or dialect? I didn't see this use case so much, for it looks just very weird for me...

Also possible because I'm getting sleepy :)
Anyways, I removed the extra newlines, problem solved.

> 
> BTW, as an off-topic
>  We should beautify the entire Timidity as original author nobody cares about it, and we develop an own independent branch of it (for example, using the "Artistic Style" program, configured for the SDL style. I use this tool to quickly beautify code fragments into a specific style, very convenient!) because it looks ugly because of lots reasons:
> 
>     * Mixed tabs and spaces indention, nobody suggests using this
> 
>     * Trailing spaces, lots of trailing spaces, they just annoying and being ballast at files

I'm not very fond of the idea, it would be an extra effort to keep things
in sync between projects (the same code, practically, is in [SDL_sound](https://github.com/icculus/SDL_sound),
[libtimidity](https://github.com/sezero/libtimidity), and who know how many other projects out there.

Wohlstand: > keep things in sync between projects (the same code, practically, is in SDL_sound, libtimidity, and who know how many other projects out there.

That means the same `astyle` conversion is needed to apply to all these cases at once. This should really avoid the future annoyance from the ugly code look, and unify the look of all these cases. Especially if develop something new, extend the functionality, etc. The major annoyance comes from code editors that are configured to work with one style, and whoops, here is alien Timidity that needs extra pain to maintain.

sezero: Well, the only good things that can be done with current v0.2-based
timidity are:
- Find a way to contact Tuukka Toivonen, and convince him to change
  his Artistic License option to [Clarified Artistic License](https://spdx.org/licenses/ClArtistic.html) which is,
  unlike the former, recognized as a free license by FSF.
- Complete the timidity-0.2 soundfont patch into a completely usable
  state (see the 'soundfont' branch in [libtimidity](https://github.com/sezero/libtimidity) - it currently lacks
  timidity.cfg machinery recognized by timidity++ e.g. concurrent use
  of gus patches and soundfonts, and a memory safety audit.)

Wohlstand: > Find a way to contact Tuukka Toivonen

The main question is, is he alive? Because of last his appearance at Timidity was at the 1995 year, 26 years ago. That means, his life would majorly be changed because of various reasons. Also, there are some people who would officially change their real name for any reason, lose all past contacts, etc. However, there is luck, because there is a small example of the case when I had successfully contacted one Chinese guy who developed some software more than 20 years ago (it's related to Sega Megadrive development), and after a year of waiting, he finally found his source code that I took for my experiments and OPN2 Bank Editor development (I successfully added support for his GIN bank format into my project).

> change his Artistic License option to Clarified Artistic License

I'm not familiar what the problem of the original license, but I would be curious to learn that...

> Complete the timidity-0.2 soundfont patch

Would be cool, yeah.

I'll append, that I would make the sort of real-time MIDI interface to use custom MIDI playing sequencers (for example, [I do have one](https://github.com/Wohlstand/BW_Midi_Sequencer) that supports loops and lots of other features) or just receive MIDI commands from the hardware/software interface, etc.

sezero: > > Find a way to contact Tuukka Toivonen
> 
> The main question is, is he alive?

Hell if I know...

>  However, there is luck,

If you can, please do let me know!

> > change his Artistic License option to Clarified Artistic License
> 
> I'm not familiar what the problem of the original license, but I would be curious to learn that...

https://en.wikipedia.org/wiki/Artistic_License#Artistic_License_1.0

> 
> > Complete the timidity-0.2 soundfont patch
> 
> Would be cool, yeah.

Indeed

> I'll append, that I would make the sort of real-time MIDI interface to use custom MIDI playing sequencers

Please not another _monster_ like fluidsynth..

Wohlstand: > Please not another monster like fluidsynth..

Don't worry, the real-time MIDI interface is very simple, it doesn't need any major changes, that's very simple because we will just re-use existing internally to provide new functionality, I already made the same at libADLMIDI and libOPNMIDI: these MIDI synthesizers are full of functionality, but still being lightweight. The main reason FluidSynth is a huge monster is that it doesn't separate unnecessary parts like the audio output interfaces and other less-significant parts. That is why the FluidLite project exists. But the fact, FluidLite strips away the MIDI file player at all, it has only a real-time MIDI interface only and the PCM output.

So, what I meant generally:
- Add the set of public calls to execute MIDI commands (note on, note off, controller change, pitch bend, program change, channel aftertouch, SysEx, etc.)
- Have the audio PCM output call to play the current state only (what MIDI commands had set, that will be played). 
- And so, how external MIDI sequencer libraries works:
  - They execute MIDI commands of the current time moment at the MIDI file position.
  - Then, they let the PCM generator generate the gap of PCM playback that has the same length as the inter-commands .delay. Then, execute the next MIDI commands.
 
The same thing works at VLC, at GZDoom, and at other things.

Issue #353, opened 2021-12-15
[suggestion] position-effects


pionere:
This is just a suggestion which I think worth discussing.
At the moment position_args are preserved as volatiles in effect_position.c . This means as far as I know the loops in _Eff_position_* function can not be optimized at all. I think it would be better to have a local copy of the values and apply it opening the possibility for SIMD operations.
I don't see a reason why would it be desired to apply different values to different part of the chunk anyway.

PR #357, opened 2021-12-16
do not use volatiles in (possible) long loops of _Eff_position_*

pionere:
#353


sezero:
@slouken: is this good?

sezero:
Well, it looks sane to me and applying this.

PR #363, opened 2021-12-25
fix handling of SDL_RWseek in music_wav.c

pionere:
- fix 'improve ParseFMT': SDL_RWseek returns -1 not 0 on error
- propagate error when SDL_RWseek returns -1
- always check the result of SDL_RWseek againts -1

### DIFF
--- a/src/codecs/music_wav.c
+++ b/src/codecs/music_wav.c
@@ -243,7 +243,7 @@ static int WAV_Play(void *context, int play_count)
         loop->current_play_count = loop->initial_play_count;
     }
     music->play_count = play_count;
-    if (SDL_RWseek(music->src, music->start, RW_SEEK_SET) < 0) {
+    if (SDL_RWseek(music->src, music->start, RW_SEEK_SET) == -1) {
         return -1;
     }
     return 0;
@@ -547,7 +547,8 @@ static int WAV_GetSome(void *context, void *data, int bytes, SDL_bool *done)
             if (loop->current_play_count > 0) {
                 --loop->current_play_count;
             }
-            SDL_RWseek(music->src, loop_start, RW_SEEK_SET);
+            if (SDL_RWseek(music->src, loop_start, RW_SEEK_SET) == -1)
+                return -1;
             looped = SDL_TRUE;
         }
     }
@@ -586,7 +587,8 @@ static int WAV_Seek(void *context, double position)
     destpos -= dest_offset % sample_size;
     if (destpos > music->stop)
         return -1;
-    SDL_RWseek(music->src, destpos, RW_SEEK_SET);
+    if (SDL_RWseek(music->src, destpos, RW_SEEK_SET) == -1)
+        return -1;
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


sezero:
>     * always check the result of SDL_RWseek againts -1

Something wrong with the negative return (`< 0`) check?

pionere:
> Something wrong with the negative return (`< 0`) check?

Just checked the implementation in SDL2 and it is completely inconsistent... So actually nothing wrong per se, but based on the description, it might return a negative value in case of overflow.
If you prefer, I or you could change it to check against < 0. Just wanted to be consistent.

sezero:
Well, it is very usual to simply check negative returns from SDL_RWseek instead of specifically -1, I guess no harm keeping it that way.

src/codecs/music_wav.c
```diff
@@ -1072,7 +1076,7 @@ static SDL_bool LoadAIFFMusic(WAV_Music *wave)
             /* Unknown/unsupported chunk: we just skip over */
             break;
         }
-    } while (next_chunk < file_length && SDL_RWseek(src, next_chunk, RW_SEEK_SET) != -1);
+    } while (next_chunk < file_length && SDL_RWseek(src, next_chunk, RW_SEEK_SET) >= 0);
```
sezero, line 1079: this one can simply be `== 0`

src/codecs/music_wav.c
```diff
@@ -651,7 +653,7 @@ static SDL_bool ParseFMT(WAV_Music *wave, Uint32 chunk_length)
         return SDL_FALSE;
     }
     chunk_length -= size;
-    if (chunk_length != 0 && !SDL_RWseek(wave->src, chunk_length, RW_SEEK_CUR)) {
+    if (chunk_length != 0 && SDL_RWseek(wave->src, chunk_length, RW_SEEK_CUR) < 0) {
```
sezero, line 656: how did this not cause an issue before?? (no one noticed? @Wohlstand?)

Wohlstand: Because this case fails when file got cutted up at middle of the chunk, but right, the failure check was wrong as I see

pionere: it is a bug in the recent change 'improve ParseFMT'

src/codecs/music_wav.c
pionere: ? I don't think that would be correct

src/codecs/music_wav.c
Wohlstand: No, this is a very old mistake made by @slouken 7 years ago:
Снимок экрана от 2021-12-25 17-15-20

Wohlstand: @pionere, you just weren't noticed that there was a bug, SDL_RWseek returns -1 when failed to seek or 0+ value that means the final offset that got been reached.

src/codecs/music_wav.c
sezero: Oh hell, my mistake, SDL_RWseek obviously follows lseek, not fseek.

PR #370, opened 2022-01-25
music_mpg123.c: Fixed memory access after free

Wohlstand:
Accessing the `music->handle` after MPEG123_Delete(music) will lead to crash


sezero:
The patch is incomplete: lines 294 and 301 need the same change.

sezero:
Thanks.

PR #376, opened 2022-04-10
Compilation fix for OS X/PowerPC

cellularmitosis:
This PR fixes a compilation error on OS X Tiger (PowerPC).

Based on the surrounding code, I'd say it is likely that someone refactored `song->` to `retval->` but forgot to update the PowerPC case to match, but didn't encounter the build failure because they weren't on a PowerPC machine.


sezero:
Thanks!

PR #378, opened 2022-04-18
Add GME for video game music support

connorjclark:
This is a partial merge of the work by @Wohlstand referenced in #233.

- ~Grabs the latest GME from https://bitbucket.org/mpyne/game-music-emu/src/master/ (commit `b3d158`)~
- Grabs music_gme.c and edits to music.c from https://github.com/WohlSoft/SDL-Mixer-X (commit `bed335`)
- Modifies `music_gme.c` to use a new `GME_StartTrack`, which is simpler to work with than the "process string args" approach that exists in the fork. With this, a single `Mix_Music` handle can be used and its track changed at any time
- Modifies `music_gme.c` to better match the code style in other files (probably could use more work here)

This is working for me when compiling with Emscripten and running in the browser. I verified these formats work:
- gbs
- nsf
- spc
- vgm

I don't have an `m3u`. There are probably other formats, but I couldn't find a list of what GME supports.

Unfortunately I'm not sure how to verify this would work on other platforms, hopefully someone can help out there. Particularly, I have no idea how to edit the `configure` file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -571,6 +571,52 @@ if test x$enable_music_midi_timidity != xyes -a \
     AC_MSG_WARN([MIDI support disabled])
 fi
 
+AC_ARG_ENABLE([music-gme],
+[AS_HELP_STRING([--enable-music-gme], [enable Game Music Emu support [default=no]])],
+              [], [enable_music_gme=no])
+AC_ARG_ENABLE([music-gme-shared],
+[AS_HELP_STRING([--enable-music-gme-shared], [dynamically load libgme library [default=yes]])],
+              [], [enable_music_gme_shared=yes])
+if test x$enable_music_gme = xyes; then
+    PKG_CHECK_MODULES([GME], [libgme], [dnl
+            have_gme_hdr=yes
+            have_gme_lib=yes
+            have_gme_pc=yes
+        ], [dnl
+            AC_CHECK_HEADER([gme/gme.h], [have_gme_hdr=yes])
+            AC_CHECK_LIB([gme], [gme_open_data], [have_gme_lib=yes;GME_LIBS="-lgme"])
+        ])
+    if test x$have_gme_hdr = xyes -a x$have_gme_lib = xyes; then
+        have_gme=yes
+        case "$host" in
+            *-*-darwin*)
+                gme_lib=[`find_lib libgme.dylib`]
+                ;;
+            *-*-cygwin* | *-*-mingw*)
+                gme_lib=[`find_lib "libgme*.dll"`]
+                ;;
+            *)
+                gme_lib=[`find_lib "libgme.so.*"`]
+                ;;
+        esac
+        EXTRA_CFLAGS="$EXTRA_CFLAGS -DMUSIC_GME $GME_CFLAGS"
+        if test x$enable_music_gme_shared = xyes && test x$gme_lib != x; then
+            echo "-- dynamic libgme -> $gme_lib"
+            EXTRA_CFLAGS="$EXTRA_CFLAGS -DGME_DYNAMIC=\\\"$gme_lib\\\""
+        else
+            EXTRA_LDFLAGS="$EXTRA_LDFLAGS $GME_LIBS"
+            if test x$have_gme_pc = xyes; then
+                PC_REQUIRES="$PC_REQUIRES libgme"
+            else
+                PC_LIBS="$PC_LIBS $GME_LIBS"
+            fi
+        fi
+    else
+        AC_MSG_WARN([*** Unable to find GME library (https://bitbucket.org/mpyne/game-music-emu)])
+        AC_MSG_WARN([Game Music Emu support disabled.])
+    fi
+fi
+
 AC_ARG_ENABLE([music-ogg],
 [AS_HELP_STRING([--enable-music-ogg], [enable Ogg Vorbis music [default=yes]])],
               [], [enable_music_ogg=yes])
--- a/include/SDL_mixer.h
+++ b/include/SDL_mixer.h
@@ -262,7 +262,8 @@ typedef enum {
     MUS_FLAC,
     MUS_MODPLUG_UNUSED,
     MUS_OPUS,
-    MUS_WAVPACK
+    MUS_WAVPACK,
+    MUS_GME
 } Mix_MusicType;
 
 /**
@@ -2424,6 +2425,10 @@ extern DECLSPEC int SDLCALL Mix_PausedMusic(void);
  * \since This function is available since SDL_mixer 2.6.0.
  */
 extern DECLSPEC int SDLCALL Mix_ModMusicJumpToOrder(int order);
+
+/* Tracks */
+extern DECLSPEC int SDLCALL Mix_StartTrack(Mix_Music *music, int track);
+extern DECLSPEC int SDLCALL Mix_GetNumTracks(Mix_Music *music);
 
 /**
  * Set the current position in the music stream, in seconds.
--- a/src/codecs/music_cmd.c
+++ b/src/codecs/music_cmd.c
@@ -293,6 +293,8 @@ Mix_MusicInterface Mix_MusicInterface_CMD =
     NULL,   /* LoopEnd */
     NULL,   /* LoopLength */
     NULL,   /* GetMetaTag */
+    NULL,   /* GetNumTracks */
+    NULL,   /* StartTrack */
     MusicCMD_Pause,
     MusicCMD_Resume,
     MusicCMD_Stop,
--- a/src/codecs/music_gme.c
+++ b/src/codecs/music_gme.c
@@ -1,0 +1,452 @@
+/*
+  SDL_mixer:  An audio mixer library based on the SDL library
+  Copyright (C) 1997-2022 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#ifdef MUSIC_GME
+
+#include "SDL_loadso.h"
+
+#include "music_gme.h"
+
+#include <gme/gme.h>
+
+typedef struct {
+    int loaded;
+    void *handle;
+
+    gme_err_t (*gme_open_data)(void const* data, long size, Music_Emu** out, int sample_rate);
+    int (*gme_track_count)(Music_Emu const*);
+    gme_err_t (*gme_start_track)(Music_Emu*, int index);
+    int (*gme_track_ended)(Music_Emu const*);
+    void (*gme_set_tempo)(Music_Emu*, double tempo);
+    int (*gme_voice_count)(Music_Emu const*);
+    void (*gme_mute_voice)(Music_Emu*, int index, int mute);
+#if GME_VERSION >= 0x000700
+    void (*gme_set_fade)(Music_Emu*, int start_msec, int fade_msec);
+#else
+    void (*gme_set_fade)(Music_Emu*, int start_msec);
+#endif
+    void (*gme_set_autoload_playback_limit)(Music_Emu*, int do_autoload_limit);
+    gme_err_t (*gme_track_info)(Music_Emu const*, gme_info_t** out, int track);
+    void (*gme_free_info)(gme_info_t*);
+    gme_err_t (*gme_seek)(Music_Emu*, int msec);
+    int (*gme_tell)(Music_Emu const*);
+    gme_err_t (*gme_play)(Music_Emu*, int count, short out[]);
+    void (*gme_delete)(Music_Emu*);
+} gme_loader;
+
+static gme_loader gme;
+
+#ifdef GME_DYNAMIC
+#define FUNCTION_LOADER(FUNC, SIG) \
+    gme.FUNC = (SIG) SDL_LoadFunction(gme.handle, #FUNC); \
+    if (gme.FUNC == NULL) { SDL_UnloadObject(gme.handle); return -1; }
+#else
+#define FUNCTION_LOADER(FUNC, SIG) \
+    gme.FUNC = FUNC; \
+    if (gme.FUNC == NULL) { Mix_SetError("Missing GME.framework"); return -1; }
+#endif
+
+static int GME_Load(void)
+{
+    if (gme.loaded == 0) {
+#ifdef GME_DYNAMIC
+        gme.handle = SDL_LoadObject(GME_DYNAMIC);
+        if (gme.handle == NULL) {
+            return -1;
+        }
+#endif
+        FUNCTION_LOADER(gme_open_data, gme_err_t (*)(void const*,long,Music_Emu**,int))
+        FUNCTION_LOADER(gme_track_count, int (*)(Music_Emu const*))
+        FUNCTION_LOADER(gme_start_track, gme_err_t (*)( Music_Emu*,int))
+        FUNCTION_LOADER(gme_track_ended, int (*)( Music_Emu const*))
+        FUNCTION_LOADER(gme_set_tempo, void (*)(Music_Emu*,double))
+        FUNCTION_LOADER(gme_voice_count, int (*)(Music_Emu const*))
+        FUNCTION_LOADER(gme_mute_voice, void (*)(Music_Emu*,int,int))
+#if GME_VERSION >= 0x000700
+        FUNCTION_LOADER(gme_set_fade, void (*)(Music_Emu*,int,int))
+#else
+        FUNCTION_LOADER(gme_set_fade, void (*)(Music_Emu*,int))
+#endif
+        FUNCTION_LOADER(gme_track_info, gme_err_t (*)(Music_Emu const*, gme_info_t**, int))
+        FUNCTION_LOADER(gme_free_info, void (*)(gme_info_t*))
+        FUNCTION_LOADER(gme_seek, gme_err_t (*)(Music_Emu*,int))
+        FUNCTION_LOADER(gme_tell, int (*)(Music_Emu const*))
+        FUNCTION_LOADER(gme_play, gme_err_t (*)(Music_Emu*, int, short[]))
+        FUNCTION_LOADER(gme_delete, void (*)(Music_Emu*))
+#if defined(GME_DYNAMIC)
+        gme.gme_set_autoload_playback_limit = (void (*)(Music_Emu*,int)) SDL_LoadFunction(gme.handle, "gme_set_autoload_playback_limit");
+#elif (GME_VERSION >= 0x000603)
+        gme.gme_set_autoload_playback_limit = gme_set_autoload_playback_limit;
+#else
+        gme.gme_set_autoload_playback_limit = NULL;
+#endif
+    }
+    ++gme.loaded;
+
+    return 0;
+}
+
+static void GME_Unload(void)
+{
+    if (gme.loaded == 0) {
+        return;
+    }
+    if (gme.loaded == 1) {
+#ifdef GME_DYNAMIC
+        SDL_UnloadObject(gme.handle);
+#endif
+    }
+    --gme.loaded;
+}
+
+/* This file supports Game Music Emulator music streams */
+typedef struct
+{
+    int play_count;
+    Music_Emu* game_emu;
+    int freesrc;
+    SDL_bool has_track_length;
+    int track_length;
+    int intro_length;
+    int loop_length;
+    int volume;
+    double tempo;
+    double gain;
+    SDL_AudioStream *stream;
+    void *buffer;
+    size_t buffer_size;
+    Mix_MusicMetaTags tags;
+} GME_Music;
+
+static void GME_Delete(void *context);
+
+/* Set the volume for a GME stream */
+static void GME_SetVolume(void *music_p, int volume)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    double v = SDL_floor(((double)(volume) * music->gain) + 0.5);
+    music->volume = (int)v;
+}
+
+/* Get the volume for a GME stream */
+static int GME_GetVolume(void *music_p)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    double v = SDL_floor(((double)(music->volume) / music->gain) + 0.5);
+    return (int)v;
+}
+
+static int initialize_from_track_info(GME_Music *music, int track)
+{
+    gme_info_t *musInfo;
+    SDL_bool has_loop_length = SDL_TRUE;
+    const char *err;
+
+    err = gme.gme_track_info(music->game_emu, &musInfo, track);
+    if (err != 0) {
+        Mix_SetError("GME: %s", err);
+        return -1;
+    }
+
+    music->track_length = musInfo->length;
+    music->intro_length = musInfo->intro_length;
+    music->loop_length = musInfo->loop_length;
+
+    music->has_track_length = SDL_TRUE;
+    if (music->track_length <= 0 ) {
+        music->track_length = (int)(2.5 * 60 * 1000);
+        music->has_track_length = SDL_FALSE;
+    }
+
+    if (music->intro_length < 0 ) {
+        music->intro_length = 0;
+    }
+    if (music->loop_length <= 0 ) {
+        if (music->track_length > 0) {
+            music->loop_length = music->track_length;
+        } else {
+            music->loop_length = (int)(2.5 * 60 * 1000);
+        }
+        has_loop_length = SDL_FALSE;
+    }
+
+    if (!music->has_track_length && has_loop_length) {
+        music->track_length = music->intro_length + music->loop_length;
+        music->has_track_length = SDL_TRUE;
+    }
+
+    meta_tags_set(&music->tags, MIX_META_TITLE, musInfo->song);
+    meta_tags_set(&music->tags, MIX_META_ARTIST, musInfo->author);
+    meta_tags_set(&music->tags, MIX_META_ALBUM, musInfo->game);
+    meta_tags_set(&music->tags, MIX_META_COPYRIGHT, musInfo->copyright);
+    gme.gme_free_info(musInfo);
+
+    return 0;
+}
+
+static void *GME_CreateFromRW(struct SDL_RWops *src, int freesrc)
+{
+    void *mem = 0;
+    size_t size;
+    GME_Music *music;
+    const char *err;
+
+    if (src == NULL) {
+        Mix_SetError("GME: Empty source given");
+        return NULL;
+    }
+
+    music = (GME_Music *)SDL_calloc(1, sizeof(GME_Music));
+
+    music->tempo = 1.0;
+    music->gain = 1.0;
+
+    music->stream = SDL_NewAudioStream(AUDIO_S16SYS, 2, music_spec.freq,
+                                       music_spec.format, music_spec.channels, music_spec.freq);
+    if (!music->stream) {
+        GME_Delete(music);
+        return NULL;
+    }
+
+    music->buffer_size = music_spec.samples * sizeof(Sint16) * 2/*channels*/ * music_spec.channels;
+    music->buffer = SDL_malloc(music->buffer_size);
+    if (!music->buffer) {
+        SDL_OutOfMemory();
+        GME_Delete(music);
+        return NULL;
+    }
+
+    SDL_RWseek(src, 0, RW_SEEK_SET);
+    mem = SDL_LoadFile_RW(src, &size, SDL_FALSE);
+    if (mem) {
+        err = gme.gme_open_data(mem, size, &music->game_emu, music_spec.freq);
+        SDL_free(mem);
+        if (err != 0) {
+            GME_Delete(music);
+            Mix_SetError("GME: %s", err);
+            return NULL;
+        }
+    } else {
+        SDL_OutOfMemory();
+        GME_Delete(music);
+        return NULL;
+    }
+
+    /* Set this flag BEFORE calling the gme_start_track() to fix an inability to loop forever */
+    if (gme.gme_set_autoload_playback_limit) {
+        gme.gme_set_autoload_playback_limit(music->game_emu, 0);
+    }
+
+    err = gme.gme_start_track(music->game_emu, 0);
+    if (err != 0) {
+        GME_Delete(music);
+        Mix_SetError("GME: %s", err);
+        return NULL;
+    }
+
+    gme.gme_set_tempo(music->game_emu, music->tempo);
+
+    music->volume = MIX_MAX_VOLUME;
+
+    meta_tags_init(&music->tags);
+    if (initialize_from_track_info(music, 0) == -1) {
+        GME_Delete(music);
+        return NULL;
+    }
+
+    music->freesrc = freesrc;
+    return music;
+}
+
+/* Start playback of a given Game Music Emulators stream */
+static int GME_Play(void *music_p, int play_count)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    int fade_start;
+    if (music) {
+        SDL_AudioStreamClear(music->stream);
+        music->play_count = play_count;
+        fade_start = play_count > 0 ? music->intro_length + (music->loop_length * play_count) : -1;
+#if GME_VERSION >= 0x000700
+        gme.gme_set_fade(music->game_emu, fade_start, 8000);
+#else
+        gme.gme_set_fade(music->game_emu, fade_start);
+#endif
+        gme.gme_seek(music->game_emu, 0);
+    }
+    return 0;
+}
+
+static int GME_GetSome(void *context, void *data, int bytes, SDL_bool *done)
+{
+    GME_Music *music = (GME_Music*)context;
+    int filled;
+    const char *err = NULL;
+
+    filled = SDL_AudioStreamGet(music->stream, data, bytes);
+    if (filled != 0) {
+        return filled;
+    }
+
+    if (gme.gme_track_ended(music->game_emu)) {
+        /* All done */
+        *done = SDL_TRUE;
+        return 0;
+    }
+
+    err = gme.gme_play(music->game_emu, (music->buffer_size / 2), (short*)music->buffer);
+    if (err != NULL) {
+        Mix_SetError("GME: %s", err);
+        return 0;
+    }
+
+    if (SDL_AudioStreamPut(music->stream, music->buffer, music->buffer_size) < 0) {
+        return -1;
+    }
+    return 0;
+}
+
+/* Play some of a stream previously started with GME_Play() */
+static int GME_PlayAudio(void *music_p, void *data, int bytes)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    return music_pcm_getaudio(music_p, data, bytes, music->volume, GME_GetSome);
+}
+
+/* Close the given Game Music Emulators stream */
+static void GME_Delete(void *context)
+{
+    GME_Music *music = (GME_Music*)context;
+    if (music) {
+        meta_tags_clear(&music->tags);
+        if (music->game_emu && music->freesrc) {
+            gme.gme_delete(music->game_emu);
+            music->game_emu = NULL;
+        }
+        if (music->stream) {
+            SDL_FreeAudioStream(music->stream);
+        }
+        if (music->buffer) {
+            SDL_free(music->buffer);
+        }
+        SDL_free(music);
+    }
+}
+
+// TODO: this should accept a track number, not assume the current track!
+static const char* GME_GetMetaTag(void *context, Mix_MusicMetaTag tag_type)
+{
+    GME_Music *music = (GME_Music *)context;
+    return meta_tags_get(&music->tags, tag_type);
+}
+
+/* Jump (seek) to a given position (time is in seconds) */
+static int GME_Seek(void *music_p, double time)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    gme.gme_seek(music->game_emu, (int)(SDL_floor((time * 1000.0) + 0.5)));
+    return 0;
+}
+
+static double GME_Tell(void *music_p)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    return (double)(gme.gme_tell(music->game_emu)) / 1000.0;
+}
+
+static double GME_Duration(void *music_p)
+{
+    GME_Music *music = (GME_Music*)music_p;
+    if (music->has_track_length) {
+        return (double)(music->track_length) / 1000.0;
+    } else {
+
+        return -1.0;
+    }
+}
+
+static int GME_GetNumTracks(void *music_p)
+{
+    GME_Music *music = (GME_Music *)music_p;
+    return gme.gme_track_count(music->game_emu);
+}
+
+static int GME_StartTrack(void *music_p, int track)
+{
+    GME_Music *music = (GME_Music *)music_p;
+    const char *err;
+
+    if ((track < 0) || (track >= gme.gme_track_count(music->game_emu))) {
+        track = gme.gme_track_count(music->game_emu) - 1;
+    }
+
+    err = gme.gme_start_track(music->game_emu, track);
+    if (err != 0) {
+        Mix_SetError("GME: %s", err);
+        return -1;
+    }
+
+    GME_Play(music, music->play_count);
+
+    if (initialize_from_track_info(music, track) == -1) {
+        return -1;
+    }
+
+    return 0;
+}
+
+
+Mix_MusicInterface Mix_MusicInterface_GME =
+{
+    "GME",
+    MIX_MUSIC_GME,
+    MUS_GME,
+    SDL_FALSE,
+    SDL_FALSE,
+
+    GME_Load,
+    NULL,   /* Open */
+    GME_CreateFromRW,
+    NULL,   /* CreateFromFile */
+    GME_SetVolume,
+    GME_GetVolume,
+    GME_Play,
+    NULL,   /* IsPlaying */
+    GME_PlayAudio,
+    NULL,   /* Jump */
+    GME_Seek,
+    GME_Tell,
+    GME_Duration,
+    NULL,
+    NULL,
+    NULL,
+    GME_GetMetaTag,
+    GME_GetNumTracks,
+    GME_StartTrack,
+    NULL,   /* Pause */
+    NULL,   /* Resume */
+    NULL,   /* Stop */
+    GME_Delete,
+    NULL,   /* Close */
+    GME_Unload
+};
+
+#endif /* MUSIC_GME */
--- a/src/codecs/music_gme.h
+++ b/src/codecs/music_gme.h
@@ -1,0 +1,26 @@
+/*
+  SDL_mixer:  An audio mixer library based on the SDL library
+  Copyright (C) 1997-2020 Sam Lantinga <slouken@libsdl.org>
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+/* This file supports playing chiptune files with libGME */
+
+#include "music.h"
+
+extern Mix_MusicInterface Mix_MusicInterface_GME;
--- a/src/music.c
+++ b/src/music.c
@@ -40,6 +40,7 @@
 #include "music_drflac.h"
 #include "music_flac.h"
 #include "music_wavpack.h"
+#include "music_gme.h"
 #include "native_midi/native_midi.h"
 
 #include "utils.h"
@@ -200,6 +201,9 @@ static Mix_MusicInterface *s_music_interfaces[] =
 #endif
 #ifdef MUSIC_MID_NATIVE
     &Mix_MusicInterface_NATIVEMIDI,
+#endif
+#ifdef MUSIC_GME
+    &Mix_MusicInterface_GME,
 #endif
 };
 
@@ -593,6 +597,32 @@ Mix_MusicType detect_music_type(SDL_RWops *src)
         return MUS_MP3;
     }
 
+    /* GME Specific files */
+    if (SDL_memcmp(magic, "ZXAY", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "GBS\x01", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "GYMX", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "HESM", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "KSCC", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "KSSX", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "NESM", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "NSFE", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "SAP\x0D", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "SNES", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "Vgm ", 4) == 0)
+        return MUS_GME;
+    if (SDL_memcmp(magic, "\x1f\x8b", 2) == 0)
+        return MUS_GME;
+
     /* Assume MOD format.
      *
      * Apparently there is no way to check if the file is really a MOD,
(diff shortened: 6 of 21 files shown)


src/codecs/music_gme.c
```diff
@@ -0,0 +1,532 @@
...
+        }
+        if (music->buffer) {
+            SDL_free(music->buffer);
+        }
+        SDL_free(music);
+    }
+}
+
+// TODO: this should accept a track number, not assume the current track!
+static const char* GME_GetMetaTag(void *context, Mix_MusicMetaTag tag_type)
```
connorjclark, line 395: Currently, if you wanted to enumerate all the tracks of a file you'd have to load each track via `StartChannel` and call this method. Instead, I'd like to propose adding a second, optional parameter `track` to the `Mix_GetMusicTitle/Author/etc` methods: if provided, it will return that track's metadata. Otherwise, it returns the currently playing track's metadata.

src/codecs/gme/Ym2612_Emu.h
```diff
@@ -0,0 +1,20 @@
+// YM2612 FM sound chip emulator interface
+
+// local edit... how to remove? build fails without one of these enabled.
+#define VGM_YM2612_GENS 1
```
connorjclark, line 4: FYI

Wohlstand:
Don't worry about M3U at all, it's Playlist format. The Mixer intended to play individual files, not playlists. Also, keep a note that the proper support for multi-track files the path arguments sub-system is needed to be ported too. I made it to allow per-song settings being written as a small appendix to the input file path, or being passes as a separated string if using specific API calls and/or memory reading based stuff.

connorjclark:
I thought it'd make the PR simpler for SDL_mixer maintainers to review without the path stuff. I'll wait for their input before either adding it back in this PR or doing it as a followup.

sezero:
Adjusting configury is the easy part - and since you asked, here it is: [patch_configure.txt](https://github.com/libsdl-org/SDL_mixer/files/8503690/patch_configure.txt)

However: I don't think that this is acceptable as it is:

1. Adding a horde of someone else's c++ code to this tree is too much.
What might possibly be acceptable would be linking to a libgme: That part
is a new feature, and @slouken and/or @icculus should decide whether
they want that feature.

2. The c++ code doesn't even compile for me as it is: it seems to have a
lot of assumptions about c++0x (I guess), and with my old g++-4.9 with
`-std=c++0x` still doesn't accept the code. Not to mention its utterly broken
debug_printf macro. So, my arguments from the first part about adding
someone else's large amount of c++ crapola repeat here.

Warnings from music_gme.c
```
../src/codecs/music_gme.c:426: warning: ‘GME_SetTempo’ defined but not used
../src/codecs/music_gme.c:437: warning: ‘GME_GetTempo’ defined but not used
../src/codecs/music_gme.c:446: warning: ‘GME_GetTracksCount’ defined but not used
../src/codecs/music_gme.c:455: warning: ‘GME_SetTrackMute’ defined but not used
```

The c++ side: Without c++0x:
```
In file included from ../src/codecs/gme/Ay_Apu.h:7:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/blargg_common.h:55:2: warning: identifier 'nullptr' is a keyword in C++11 [-Wc++0x-compat]
  void clear() { free( begin_ ); begin_ = nullptr; size_ = 0; }
  ^
../src/codecs/gme/blargg_common.h: In member function 'void blargg_vector<T>::clear()':
../src/codecs/gme/blargg_common.h:55:42: error: 'nullptr' was not declared in this scope
  void clear() { free( begin_ ); begin_ = nullptr; size_ = 0; }
                                          ^
../src/codecs/gme/blargg_common.h: In member function 'T& blargg_vector<T>::operator[](size_t) const':
../src/codecs/gme/blargg_common.h:58:22: error: there are no arguments to 'assert' that depend on a template parameter, so a declaration of 'assert' must be available [-fpermissive]
   assert( n <= size_ ); // <= to allow past-the-end value
                      ^
../src/codecs/gme/blargg_common.h:58:22: note: (if you use '-fpermissive', G++ will accept your code, but allowing the use of an undeclared name is deprecated)
In file included from ../src/codecs/gme/Ay_Apu.h:8:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/Blip_Buffer.h: At global scope:
../src/codecs/gme/Blip_Buffer.h:236:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>           (const Blip_Synth<quality, range>  &) = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h:237:73: error: expected ',' or '...' before '&&' token
  Blip_Synth<quality, range>           (      Blip_Synth<quality, range> &&) = delete;
                                                                         ^
../src/codecs/gme/Blip_Buffer.h:237:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>           (      Blip_Synth<quality, range> &&) = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h:237:79: error: invalid constructor; you probably meant 'Blip_Synth<quality, range> (const Blip_Synth<quality, range>&)'
../src/codecs/gme/Blip_Buffer.h:238:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>& operator=(const Blip_Synth<quality, range> &)  = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h: In member function 'void Blip_Synth<quality, range>::offset_resampled(blip_resampled_time_t, int, Blip_Buffer*) const':
../src/codecs/gme/Blip_Buffer.h:347:78: error: there are no arguments to 'assert' that depend on a template parameter, so a declaration of 'assert' must be available [-fpermissive]
  assert( (blip_long) (time >> BLIP_BUFFER_ACCURACY) < blip_buf->buffer_size_ );
                                                                              ^
In file included from ../src/codecs/gme/Ay_Apu.cpp:3:0:
../src/codecs/gme/Ay_Apu.h: In member function 'void Ay_Apu::osc_output(int, Blip_Buffer*)':
../src/codecs/gme/Ay_Apu.h:86:35: error: 'assert' was not declared in this scope
  assert( (unsigned) i < osc_count );
                                   ^
../src/codecs/gme/Ay_Apu.h: In member function 'void Ay_Apu::end_frame(blip_time_t)':
../src/codecs/gme/Ay_Apu.h:102:28: error: 'assert' was not declared in this scope
  assert( last_time >= time );
                            ^
../src/codecs/gme/Ay_Apu.cpp: In member function 'void Ay_Apu::write_data_(int, int)':
../src/codecs/gme/Ay_Apu.cpp:122:38: error: 'assert' was not declared in this scope
  assert( (unsigned) addr < reg_count );
                                      ^
../src/codecs/gme/Ay_Apu.cpp:127:52: warning: left operand of comma operator has no effect [-Wunused-value]
    debug_printf( "Wrote to I/O port %02X\n", (int) addr );
                                                    ^
In file included from ../src/codecs/gme/Ay_Apu.cpp:16:0:
../src/codecs/gme/Ay_Apu.cpp: In member function 'void Ay_Apu::run_until(blip_time_t)':
../src/codecs/gme/blargg_source.h:19:38: error: 'assert' was not declared in this scope
 #define require( expr ) assert( expr )
                                      ^
../src/codecs/gme/Ay_Apu.cpp:166:2: note: in expansion of macro 'require'
  require( final_end_time >= last_time );
  ^
In file included from ../src/codecs/gme/Ay_Apu.h:8:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/Blip_Buffer.h: In instantiation of 'void Blip_Synth<quality, range>::offset_resampled(blip_resampled_time_t, int, Blip_Buffer*) const [with int quality = 12; int range = 1; blip_resampled_time_t = unsigned int]':
../src/codecs/gme/Blip_Buffer.h:457:64:   required from 'void Blip_Synth<quality, range>::offset(blip_time_t, int, Blip_Buffer*) const [with int quality = 12; int range = 1; blip_time_t = int]'
../src/codecs/gme/Ay_Apu.cpp:278:51:   required from here
../src/codecs/gme/Blip_Buffer.h:347:78: error: 'assert' was not declared in this scope
  assert( (blip_long) (time >> BLIP_BUFFER_ACCURACY) < blip_buf->buffer_size_ );
                                                                              ^
make: *** [build/Ay_Apu.lo] Error 1
```

With c++0x:
```
In file included from ../src/codecs/gme/Ay_Apu.h:7:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/blargg_common.h:55:2: warning: identifier 'nullptr' is a keyword in C++11 [-Wc++0x-compat]
  void clear() { free( begin_ ); begin_ = nullptr; size_ = 0; }
  ^
../src/codecs/gme/blargg_common.h: In member function 'void blargg_vector<T>::clear()':
../src/codecs/gme/blargg_common.h:55:42: error: 'nullptr' was not declared in this scope
  void clear() { free( begin_ ); begin_ = nullptr; size_ = 0; }
                                          ^
../src/codecs/gme/blargg_common.h: In member function 'T& blargg_vector<T>::operator[](size_t) const':
../src/codecs/gme/blargg_common.h:58:22: error: there are no arguments to 'assert' that depend on a template parameter, so a declaration of 'assert' must be available [-fpermissive]
   assert( n <= size_ ); // <= to allow past-the-end value
                      ^
../src/codecs/gme/blargg_common.h:58:22: note: (if you use '-fpermissive', G++ will accept your code, but allowing the use of an undeclared name is deprecated)
In file included from ../src/codecs/gme/Ay_Apu.h:8:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/Blip_Buffer.h: At global scope:
../src/codecs/gme/Blip_Buffer.h:236:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>           (const Blip_Synth<quality, range>  &) = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h:237:73: error: expected ',' or '...' before '&&' token
  Blip_Synth<quality, range>           (      Blip_Synth<quality, range> &&) = delete;
                                                                         ^
../src/codecs/gme/Blip_Buffer.h:237:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>           (      Blip_Synth<quality, range> &&) = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h:237:79: error: invalid constructor; you probably meant 'Blip_Synth<quality, range> (const Blip_Synth<quality, range>&)'
../src/codecs/gme/Blip_Buffer.h:238:79: warning: defaulted and deleted functions only available with -std=c++11 or -std=gnu++11
  Blip_Synth<quality, range>& operator=(const Blip_Synth<quality, range> &)  = delete;
                                                                               ^
../src/codecs/gme/Blip_Buffer.h: In member function 'void Blip_Synth<quality, range>::offset_resampled(blip_resampled_time_t, int, Blip_Buffer*) const':
../src/codecs/gme/Blip_Buffer.h:347:78: error: there are no arguments to 'assert' that depend on a template parameter, so a declaration of 'assert' must be available [-fpermissive]
  assert( (blip_long) (time >> BLIP_BUFFER_ACCURACY) < blip_buf->buffer_size_ );
                                                                              ^
In file included from ../src/codecs/gme/Ay_Apu.cpp:3:0:
../src/codecs/gme/Ay_Apu.h: In member function 'void Ay_Apu::osc_output(int, Blip_Buffer*)':
../src/codecs/gme/Ay_Apu.h:86:35: error: 'assert' was not declared in this scope
  assert( (unsigned) i < osc_count );
                                   ^
../src/codecs/gme/Ay_Apu.h: In member function 'void Ay_Apu::end_frame(blip_time_t)':
../src/codecs/gme/Ay_Apu.h:102:28: error: 'assert' was not declared in this scope
  assert( last_time >= time );
                            ^
../src/codecs/gme/Ay_Apu.cpp: In member function 'void Ay_Apu::write_data_(int, int)':
../src/codecs/gme/Ay_Apu.cpp:122:38: error: 'assert' was not declared in this scope
  assert( (unsigned) addr < reg_count );
                                      ^
../src/codecs/gme/Ay_Apu.cpp:127:52: warning: left operand of comma operator has no effect [-Wunused-value]
    debug_printf( "Wrote to I/O port %02X\n", (int) addr );
                                                    ^
In file included from ../src/codecs/gme/Ay_Apu.cpp:16:0:
../src/codecs/gme/Ay_Apu.cpp: In member function 'void Ay_Apu::run_until(blip_time_t)':
../src/codecs/gme/blargg_source.h:19:38: error: 'assert' was not declared in this scope
 #define require( expr ) assert( expr )
                                      ^
../src/codecs/gme/Ay_Apu.cpp:166:2: note: in expansion of macro 'require'
  require( final_end_time >= last_time );
  ^
In file included from ../src/codecs/gme/Ay_Apu.h:8:0,
                 from ../src/codecs/gme/Ay_Apu.cpp:3:
../src/codecs/gme/Blip_Buffer.h: In instantiation of 'void Blip_Synth<quality, range>::offset_resampled(blip_resampled_time_t, int, Blip_Buffer*) const [with int quality = 12; int range = 1; blip_resampled_time_t = unsigned int]':
../src/codecs/gme/Blip_Buffer.h:457:64:   required from 'void Blip_Synth<quality, range>::offset(blip_time_t, int, Blip_Buffer*) const [with int quality = 12; int range = 1; blip_time_t = int]'
../src/codecs/gme/Ay_Apu.cpp:278:51:   required from here
../src/codecs/gme/Blip_Buffer.h:347:78: error: 'assert' was not declared in this scope
  assert( (blip_long) (time >> BLIP_BUFFER_ACCURACY) < blip_buf->buffer_size_ );
                                                                              ^
make: *** [build/Ay_Apu.lo] Error 1
```

connorjclark:
Removed the GME folder, having the user link against the library is better.

re: the unused functions for tempo, is there a preference for adding additional methods to `Mix_MusicInterface` in this PR to expose setting the speed, or as a separate PR?

slouken:
A separate PR would be best.

include/SDL_mixer.h
```diff
@@ -620,6 +621,10 @@ extern DECLSPEC int SDLCALL Mix_PausedMusic(void);
  */
 extern DECLSPEC int SDLCALL Mix_ModMusicJumpToOrder(int order);
 
+/* Tracks */
+extern DECLSPEC void SDLCALL Mix_StartTrack(int track);
+extern DECLSPEC int SDLCALL Mix_GetNumTracks(Mix_Music *music);
+
```
connorjclark, line 627: Any thoughts on these APIs/naming?

slouken: The naming seems fine.

@sezero, does this fit in with any of the other supported music formats?
@icculus, do you have any thoughts on this PR?

sezero: > @sezero, does this fit in with any of the other supported music formats?

No, I don't think so.

configure.ac
```diff
@@ -568,6 +568,19 @@ if test x$enable_music_midi_timidity != xyes -a \
     AC_MSG_WARN([MIDI support disabled])
 fi
 
+AC_ARG_ENABLE([music-gme],
```
sezero, line 571: As it is, the configury is broken: here is a patch to fix it: [patch_378b.txt](https://github.com/libsdl-org/SDL_mixer/files/8605149/patch_378b.txt)

And after that, the PR definitely needs squashing.

src/codecs/music_gme.c
```diff
@@ -0,0 +1,493 @@
+// https://github.com/WohlSoft/SDL-Mixer-X
```
sezero, line 1: Lose unnecessary comment

src/codecs/music_gme.h
```diff
@@ -0,0 +1,35 @@
...
+/* This file supports playing chiptune files with libGME */
+
+#include "music.h"
+
+extern Mix_MusicInterface Mix_MusicInterface_GME;
+
+#ifdef MUSIC_GME
+
+extern void _Mix_GME_SetSpcEchoDisabled(void *music, int disabled);
+extern int  _Mix_GME_GetSpcEchoDisabled(void *music);
```
sezero, line 33: AFAICS, not referenced anywhere. What are these for?

src/codecs/music_gme.c
```diff
@@ -0,0 +1,493 @@
...
+        ret = music->echo_disabled;
+    }
+
+    return ret;
+}
+
+static void GME_Delete(void *context);
+
+/* Set the volume for a GME stream */
+void GME_SetVolume(void *music_p, int volume)
```
sezero, line 182: private procedures should be made static

src/codecs/music_gme.c
```diff
@@ -0,0 +1,493 @@
...
+{
+    GME_Music *music = (GME_Music*)music_p;
+    double v = SDL_floor(((double)(volume) * music->gain) + 0.5);
+    music->volume = (int)v;
+}
+
+/* Get the volume for a GME stream */
+int GME_GetVolume(void *music_p)
```
sezero, line 190: same

src/codecs/music_gme.c
```diff
@@ -0,0 +1,493 @@
...
+{
+    GME_Music *music = (GME_Music*)music_p;
+    double v = SDL_floor(((double)(music->volume) / music->gain) + 0.5);
+    return (int)v;
+}
+
+int initialize_from_track_info(GME_Music *music, int track)
```
sezero, line 197: same

src/codecs/music_gme.h
Wohlstand: This is one of the most recent features I added to GME, but the original author seems away for any reason and he won't be able to review [my pull requests](https://bitbucket.org/mpyne/game-music-emu/pull-requests/) for a long time...

One note: at my CMake build there is a check for the presence of certain API functions at currently available GME assembly, and so, if using an old version, this function becomes a dummy.

src/codecs/music_gme.c
```diff
@@ -0,0 +1,493 @@
...
+    GME_Load,
+    NULL,   /* Open */
+    GME_CreateFromRW,
+    NULL,   /* CreateFromFile */
+    GME_SetVolume,
+    GME_GetVolume,
+    GME_Play,
+    NULL,   /* IsPlaying */
+    GME_PlayAudio,
+    NULL,       /* Jump */
```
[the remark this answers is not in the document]
Wohlstand: There are time-based seek calls used, the Jump is too specific to tracker music, and chiptunes are much different from them (they are assembly running on a certain processor chip, or sequence of raw register commands for the state-based sound generator chip, etc.)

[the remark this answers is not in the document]
sezero: I see, OK,

src/codecs/music_gme.h
Wohlstand: > specific to your fork 

yet, until original author will accept my pull-requests

Wohlstand: @sezero, two of these forks you mentioned, are very outdated (@kode54, author of one of them, can confirm). Also, you may see, that their code is totally different from the original in the case of repo organizing, way of building, internality, etc. Changes I did at my copy, are intended to be merged into mainstream, but they didn't because the current maintainer is not available. So, from here you may drop (or comment out until the thing gets merged into the mainstream once upon)

src/codecs/music_gme.c
Wohlstand: Thanks for the notice, I'll mark these calls being static at my fork side...

Wohlstand:
P.S. Late reply to @sezero about GME compilation: it requires the C++11 support by the compiler side, and elder GCCs obviously lack the support for it. The `-std=c++11` is mainly used to build the thing. Also, using CMake build provided by the mainstream version, or my custom minified CMake build at my AudioCodecs aggregation repo. I don't remind, but something worked at GCC 4.8 or not, I do remind about my other C++11 project, not speaking about GME that probably also failed, and I installed the system-wide version from packages or something also...

sezero:
> P.S. Late reply to @sezero about GME compilation: it requires the C++11

The mainstream code does compile for me using gcc4.9 (its cmake specifically adds -std=c++11)

configure.ac
```diff
@@ -569,6 +569,58 @@ if test x$enable_music_midi_timidity != xyes -a \
     AC_MSG_WARN([MIDI support disabled])
 fi
 
+if test x$enable_music_midi_timidity != xyes -a \
+        x$use_music_midi_native != xyes -a x$use_music_midi_native != xyes_cpp -a \
```
sezero, line 573: The timidity lines somehow get duplicated: attached patch fixes that and tidies it up a bit more. [patch.txt](https://github.com/libsdl-org/SDL_mixer/files/8609463/patch.txt)

sezero:
Here are some clean-ups:

- `gme_disable_echo` is not in mainstream yet, only a pull request: remove.
- `GME_HAS_SET_AUTOLOAD_PLAYBACK_LIMIT` isn't defined anywhere: Safest way
  would be a `GME_VERSION` check. And unlike what is documented in gme.h, it
  is in 0.6.3 and newer, _not_ 0.6.2.
- Followin commit ecce55a in master, the macOS case needs revising.

Patch attached: [patch1.txt](https://github.com/libsdl-org/SDL_mixer/files/8609691/patch1.txt), and also inlined below:
```diff
diff --git a/src/codecs/music_gme.c b/src/codecs/music_gme.c
index 841a86c..3b44ef8 100644
--- a/src/codecs/music_gme.c
+++ b/src/codecs/music_gme.c
@@ -44,7 +44,6 @@ typedef struct {
     void (*gme_set_fade)(Music_Emu*, int start_msec);
 #endif
     void (*gme_set_autoload_playback_limit)(Music_Emu*, int do_autoload_limit);
-    void (*gme_disable_echo)(Music_Emu*, int disable);
     gme_err_t (*gme_track_info)(Music_Emu const*, gme_info_t** out, int track);
     void (*gme_free_info)(gme_info_t*);
     gme_err_t (*gme_seek)(Music_Emu*, int msec);
@@ -61,7 +60,8 @@ static gme_loader gme;
     if (gme.FUNC == NULL) { SDL_UnloadObject(gme.handle); return -1; }
 #else
 #define FUNCTION_LOADER(FUNC, SIG) \
-    gme.FUNC = FUNC;
+    gme.FUNC = FUNC; \
+    if (gme.FUNC == NULL) { Mix_SetError("Missing GME.framework"); return -1; }
 #endif
 
 static int GME_Load(void)
@@ -72,13 +72,6 @@ static int GME_Load(void)
         if (gme.handle == NULL) {
             return -1;
         }
-#elif defined(__MACOSX__)
-        extern gme_err_t gme_open_data(void const*,long,Music_Emu**,int) __attribute__((weak_import));
-        if (gme_open_data == NULL) {
-            /* Missing weakly linked framework */
-            Mix_SetError("Missing GME.framework");
-            return -1;
-        }
 #endif
         FUNCTION_LOADER(gme_open_data, gme_err_t (*)(void const*,long,Music_Emu**,int))
         FUNCTION_LOADER(gme_track_count, int (*)(Music_Emu const*))
@@ -100,19 +93,11 @@ static int GME_Load(void)
         FUNCTION_LOADER(gme_delete, void (*)(Music_Emu*))
 #if defined(GME_DYNAMIC)
         gme.gme_set_autoload_playback_limit = (void (*)(Music_Emu*,int)) SDL_LoadFunction(gme.handle, "gme_set_autoload_playback_limit");
-#elif defined(GME_HAS_SET_AUTOLOAD_PLAYBACK_LIMIT)
+#elif (GME_VERSION >= 0x000603)
         gme.gme_set_autoload_playback_limit = gme_set_autoload_playback_limit;
 #else
         gme.gme_set_autoload_playback_limit = NULL;
 #endif
-
-#if defined(GME_DYNAMIC)
-        gme.gme_disable_echo = (void (*)(Music_Emu*,int)) SDL_LoadFunction(gme.handle, "gme_disable_echo");
-#elif defined(GME_HAS_DISABLE_ECHO)
-        gme.gme_disable_echo = gme_disable_echo;
-#else
-        gme.gme_disable_echo = NULL;
-#endif
     }
     ++gme.loaded;
 
```

Wohlstand:
> GME_HAS_SET_AUTOLOAD_PLAYBACK_LIMIT

This is a part of the CMake-side auto-check result I wrote to detect compatible ABI, there are older GME versions that lack this call, and without this call, on newer GME some songs won't loop when expected they will loop.

sezero:
> > GME_HAS_SET_AUTOLOAD_PLAYBACK_LIMIT
> 
> This is a part of the CMake-side auto-check result I wrote to detect compatible ABI,

In any case, the `GME_VERSION` check I used above should be equivalent and enough.

> there are older GME versions that lack this call, and without this call, on newer GME some songs won't loop when expected they will loop.

Should require at least version 0.6.3 and newer then? (I guess that would be too much but...)

Wohlstand:
> Should we require at least version 0.6.3 and newer then? (I guess that would be too much but...)

There are several copies of the 0.6.3 version without this call, I know that because I long time watching for the GME side, and I met compilation errors at some machines when using GME out of my own build (taken from the mainstream repo and later)

sezero:
> There are several copies of the 0.6.3 version without this call,

Eh?? Can you give links?

>   (taken from the mainstream repo and later)

I can't even download release tarballs linked at main bitbucket site (OpenSSL errors)

Wohlstand:
> Eh?? Can you give links?

Can't remind, you if they were found at package managers or there are was middle versions of 0.6.3 while its development process

sezero:
Looked at fedora and ubuntu sources for 0.6.3, they have the gme_set_autoload_playback_limit
call, and so does the 0.6.3 tag in mainstream git, therefore the above check _should_ be good.

Wohlstand:
Some news: no need to remove SPC Echo-related usage, because my pull request got been finally merged: https://bitbucket.org/mpyne/game-music-emu/pull-requests/30, now it's the mainstream part.

sezero:
The OP (@connorjclark) seems to have lost interest?

connorjclark:
I'll try to find the time this evening to apply the suggested patches.

> Some news: no need to remove SPC Echo-related usage, because my pull request got been finally merged: [bitbucket.org/mpyne/game-music-emu/pull-requests/30](https://bitbucket.org/mpyne/game-music-emu/pull-requests/30), now it's the mainstream part.

Would you like to handle this in a new PR? I wouldn't know how to test this.

sezero:
> I'll try to find the time this evening to apply the suggested patches.

OK good.

While doing that I suggest that you drop changes to `configure` itself:
I will regenerate that for you, so only concern yourself with `configure.ac`

> > Some news: no need to remove SPC Echo-related usage, because my pull request got been finally merged: [bitbucket.org/mpyne/game-music-emu/pull-requests/30](https://bitbucket.org/mpyne/game-music-emu/pull-requests/30), now it's the mainstream part.
> 
> Would you like to handle this in a new PR? I wouldn't know how to test this.

There is no code at all using `gme_disable_echo` in here: we remove it for
the time being. If music_gme.c gets real use of it then we can think about
what to do with that.

Here are the patches that need applying (same as above but repeated here
for convenience, and with changes to generated `configure` file dropped):
1. Fixes to configure.ac: [patch-01.txt](https://github.com/libsdl-org/SDL_mixer/files/8734058/patch-01.txt)
2. Fixes / cleanups to music_gme.c (patch file has explanations): [patch-02.txt](https://github.com/libsdl-org/SDL_mixer/files/8734079/patch-02.txt)

Wohlstand:
> There is no code at all using gme_disable_echo in here: we remove it for
the time being. If music_gme.c gets real use of it then we can think about
what to do with that.

On my side I already use it: I do apply various sound effects using `Mix_SetPostMix()` call and siblings, and when I play SPC files, their built-in echo conflicts with effects I do apply on my side, they result in the mess. So, that option I introduced, allows disabling the native echo effect of SPC files, making them play in plain form. This resolves the conflict with custom sound effects of the echo and reverb family.

src/codecs/music_gme.c
```diff
@@ -0,0 +1,469 @@
...
+  3. This notice may not be removed or altered from any source distribution.
+*/
+
+#ifdef MUSIC_GME
+
+#include "SDL_loadso.h"
+
+#include "music_gme.h"
+
+#include <gme/gme.h>
```
Wohlstand, line 28: Why I used <gme.h> than <gme/gme.h>?
- In some systems with older GME versions, the header was installed at the "include" directory directly and wasn't included into sub-directory
- Since some versions, the header started to be placed at the "gme" sub-directory because new public headers got been added by Michael Pyne (for internal needs of GME itself). And so, at the configure script you should detect the presence of the "gme" directory and attach it to the include path.

src/codecs/music_gme.c
```diff
@@ -0,0 +1,469 @@
...
+
+    music->volume = MIX_MAX_VOLUME;
+
+    meta_tags_init(&music->tags);
+    if (initialize_from_track_info(music, 0) == -1) {
+        GME_Delete(music);
+        return NULL;
+    }
+
+    music->freesrc = freesrc;
```
Wohlstand, line 291: You absolutely don't need to keep the file loaded in the memory: it gets parsed once by GME itself and is no longer used. I suggest doing the same as at the music_xmp.c and at my original version of music_gme.c on the MixerX side:
```c
    if (freesrc) {
        SDL_RWclose(src);
    }
```

At my version, I do use two overlay calls, and the freeing is done at them. In this version, you need to do this at this line.

src/codecs/music_gme.c
sezero: The oldest package I found for fedora was from Aug. 2011 for Fedora-16:
http://archives.fedoraproject.org/pub/archive/fedora/linux/releases/16/Everything/i386/os/Packages/game-music-emu-devel-0.5.5-1.fc16.i686.rpm
It installs gme headers under `gme` directory, so we should be safe here

Wohlstand: Fine

sezero:
@slouken: Do we want this feature?

slouken:
I'm not opposed to it, although the official library page says that it's on maintenance life support rather than being actively maintained. Do we have examples of games that use this music format? Are there good test files that we can use to verify the functionality?

connorjclark:
The [Zelda Classic](https://www.zeldaclassic.com/) game engine ([github](https://github.com/ArmageddonGames/ZeldaClassic)) uses GME to enable its game creators to leverage tons of video game music in their own games. I recently ported this engine to the web, using SDL as the backend (and a fork of SDL_mixer with this patch to support GME).

Here's one example game using gbs music: https://hoten.cc/zc/play/?quest=623/Rite+of+the+Storm/Rite+of+the+Storm.qst . You'll see the gbs file is downloaded once the title sequence for the game loads (sorry, you'll have to wait for the page to download some things first)
image


I have access to a few dozen files in these various formats, should I add a few for test cases?

Wohlstand:
The set of test files, crafted by various people (and two of them crafted by me a while ago):
[test-spc-files.zip](https://github.com/libsdl-org/SDL_mixer/files/8743149/test-spc-files.zip)

Also, very important note, I forgot to tell you: on my side, through path arguments its possible to set the so-called "gaining factor" - its multiplier of the volume level, it allows to make too silent music sound louder and allows to calm down too loud music, making it a bit quite. Initially, I also hadn't the gaining factor, but with a time people noticed that it's VERY important to add it because there are lots of chiptunes that sound too quiet with default output. The constant gaining factor is not suitable, because, among silent files, there are opposite loud files that may cause clip-on integer output formats.

connorjclark:
Any other changes needed?

sezero:
Looking at this again, every music interface other than music_gme have
the `StartTrack` and `GetNumTracks` fields uninitialized: they should be.
Here is a patch (attached and also inlined below) which does that, and
also makes a tiny clean-up in music_gme. [patch-03.txt](https://github.com/libsdl-org/SDL_mixer/files/8924495/patch-03.txt)

```diff
diff -u a/src/codecs/music_gme.c b/src/codecs/music_gme.c
--- a/src/codecs/music_gme.c
+++ b/src/codecs/music_gme.c
@@ -1,6 +1,6 @@
 /*
   SDL_mixer:  An audio mixer library based on the SDL library
-  Copyright (C) 1997-2020 Sam Lantinga <slouken@libsdl.org>
+  Copyright (C) 1997-2022 Sam Lantinga <slouken@libsdl.org>
 
   This software is provided 'as-is', without any express or implied
   warranty.  In no event will the authors be held liable for any damages
@@ -124,7 +124,6 @@
     Music_Emu* game_emu;
     int freesrc;
     SDL_bool has_track_length;
-    int echo_disabled;
     int track_length;
     int intro_length;
     int loop_length;
@@ -432,7 +431,7 @@
     GME_Play,
     NULL,   /* IsPlaying */
     GME_PlayAudio,
-    NULL,       /* Jump */
+    NULL,   /* Jump */
     GME_Seek,
     GME_Tell,
     GME_Duration,
@@ -449,6 +448,5 @@
-
     GME_StartTrack,
-    GME_GetNumTracks,
+    GME_GetNumTracks
 };
 
-#endif /* MUSIC_GME */
\ No newline at end of file
+#endif /* MUSIC_GME */
diff --git a/src/codecs/music_cmd.c b/src/codecs/music_cmd.c
index a293f92..444f342 100644
--- a/src/codecs/music_cmd.c
+++ b/src/codecs/music_cmd.c
@@ -298,7 +298,9 @@ Mix_MusicInterface Mix_MusicInterface_CMD =
     MusicCMD_Stop,
     MusicCMD_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_CMD */
diff --git a/src/codecs/music_drflac.c b/src/codecs/music_drflac.c
index 635e975..67f6fce 100644
--- a/src/codecs/music_drflac.c
+++ b/src/codecs/music_drflac.c
@@ -414,7 +414,9 @@ Mix_MusicInterface Mix_MusicInterface_DRFLAC =
     DRFLAC_Stop,
     DRFLAC_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_FLAC_DRFLAC */
diff --git a/src/codecs/music_drmp3.c b/src/codecs/music_drmp3.c
index 7af022d..5f75f2f 100644
--- a/src/codecs/music_drmp3.c
+++ b/src/codecs/music_drmp3.c
@@ -287,7 +287,9 @@ Mix_MusicInterface Mix_MusicInterface_DRMP3 =
     DRMP3_Stop,
     DRMP3_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MP3_DRMP3 */
diff --git a/src/codecs/music_flac.c b/src/codecs/music_flac.c
index 2e19d80..60b2e54 100644
--- a/src/codecs/music_flac.c
+++ b/src/codecs/music_flac.c
@@ -748,13 +748,15 @@ Mix_MusicInterface Mix_MusicInterface_FLAC =
     FLAC_LoopStart,
     FLAC_LoopEnd,
     FLAC_LoopLength,
-    FLAC_GetMetaTag,/* GetMetaTag */
+    FLAC_GetMetaTag,
     NULL,   /* Pause */
     NULL,   /* Resume */
     FLAC_Stop,   /* Stop */
     FLAC_Delete,
     NULL,   /* Close */
-    FLAC_Unload
+    FLAC_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_FLAC_LIBFLAC */
diff --git a/src/codecs/music_fluidsynth.c b/src/codecs/music_fluidsynth.c
index 49d9699..39c633a 100644
--- a/src/codecs/music_fluidsynth.c
+++ b/src/codecs/music_fluidsynth.c
@@ -368,7 +368,9 @@ Mix_MusicInterface Mix_MusicInterface_FLUIDSYNTH =
     FLUIDSYNTH_Stop,
     FLUIDSYNTH_Delete,
     NULL,   /* Close */
-    FLUIDSYNTH_Unload
+    FLUIDSYNTH_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MID_FLUIDSYNTH */
diff --git a/src/codecs/music_modplug.c b/src/codecs/music_modplug.c
index aca5d5e..f1744eb 100644
--- a/src/codecs/music_modplug.c
+++ b/src/codecs/music_modplug.c
@@ -368,7 +368,9 @@ Mix_MusicInterface Mix_MusicInterface_MODPLUG =
     MODPLUG_Stop,
     MODPLUG_Delete,
     NULL,   /* Close */
-    MODPLUG_Unload
+    MODPLUG_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MOD_MODPLUG */
diff --git a/src/codecs/music_mpg123.c b/src/codecs/music_mpg123.c
index a8198d8..4b23c34 100644
--- a/src/codecs/music_mpg123.c
+++ b/src/codecs/music_mpg123.c
@@ -539,7 +539,9 @@ Mix_MusicInterface Mix_MusicInterface_MPG123 =
     MPG123_Stop,
     MPG123_Delete,
     MPG123_Close,
-    MPG123_Unload
+    MPG123_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MP3_MPG123 */
diff --git a/src/codecs/music_nativemidi.c b/src/codecs/music_nativemidi.c
index 22fbcc0..408881e 100644
--- a/src/codecs/music_nativemidi.c
+++ b/src/codecs/music_nativemidi.c
@@ -113,7 +113,9 @@ Mix_MusicInterface Mix_MusicInterface_NATIVEMIDI =
     NATIVEMIDI_Stop,
     NATIVEMIDI_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MID_NATIVE */
diff --git a/src/codecs/music_ogg.c b/src/codecs/music_ogg.c
index 16af47b..15688ac 100644
--- a/src/codecs/music_ogg.c
+++ b/src/codecs/music_ogg.c
@@ -548,7 +548,9 @@ Mix_MusicInterface Mix_MusicInterface_OGG =
     OGG_Stop,
     OGG_Delete,
     NULL,   /* Close */
-    OGG_Unload
+    OGG_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_OGG */
diff --git a/src/codecs/music_ogg_stb.c b/src/codecs/music_ogg_stb.c
index b33f38e..c19e523 100644
--- a/src/codecs/music_ogg_stb.c
+++ b/src/codecs/music_ogg_stb.c
@@ -477,7 +477,9 @@ Mix_MusicInterface Mix_MusicInterface_OGG =
     OGG_Stop,
     OGG_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_OGG */
diff --git a/src/codecs/music_opus.c b/src/codecs/music_opus.c
index c298095..d1048a8 100644
--- a/src/codecs/music_opus.c
+++ b/src/codecs/music_opus.c
@@ -519,7 +519,9 @@ Mix_MusicInterface Mix_MusicInterface_Opus =
     OPUS_Stop,
     OPUS_Delete,
     NULL,   /* Close */
-    OPUS_Unload
+    OPUS_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_OPUS */
diff --git a/src/codecs/music_timidity.c b/src/codecs/music_timidity.c
index 5822c4b..cdc3916 100644
--- a/src/codecs/music_timidity.c
+++ b/src/codecs/music_timidity.c
@@ -289,7 +289,9 @@ Mix_MusicInterface Mix_MusicInterface_TIMIDITY =
     TIMIDITY_Stop,
     TIMIDITY_Delete,
     TIMIDITY_Close,
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MID_TIMIDITY */
diff --git a/src/codecs/music_xmp.c b/src/codecs/music_xmp.c
index 39d46c3..057ac7f 100644
--- a/src/codecs/music_xmp.c
+++ b/src/codecs/music_xmp.c
@@ -431,7 +431,9 @@ Mix_MusicInterface Mix_MusicInterface_XMP =
     XMP_Stop,
     XMP_Delete,
     NULL,   /* Close */
-    XMP_Unload
+    XMP_Unload,
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_MOD_XMP */
diff --git a/src/codecs/music_wav.c b/src/codecs/music_wav.c
index e4140c2..11b92af 100644
--- a/src/codecs/music_wav.c
+++ b/src/codecs/music_wav.c
@@ -931,8 +931,7 @@ static SDL_bool LoadWAVMusic(WAV_Music *wave)
         if (chunk_length == 0)
             break;
 
-        switch (chunk_type)
-        {
+        switch (chunk_type) {
         case FMT:
             found_FMT = SDL_TRUE;
             if (!ParseFMT(wave, chunk_length))
@@ -1135,7 +1134,6 @@ static SDL_bool LoadAIFFMusic(WAV_Music *wave)
         return SDL_FALSE;
     }
 
-
     wave->samplesize = channels * (samplesize / 8);
     wave->stop = wave->start + channels * numsamples * (samplesize / 8);
 
@@ -1259,7 +1257,9 @@ Mix_MusicInterface Mix_MusicInterface_WAV =
     WAV_Stop, /* Stop */
     WAV_Delete,
     NULL,   /* Close */
-    NULL    /* Unload */
+    NULL,   /* Unload */
+    NULL,   /* StartTrack   */
+    NULL    /* GetNumTracks */
 };
 
 #endif /* MUSIC_WAV */
```

sezero:
OK, I have nothing further.

The source files and any necessary defines & etc will still need adding to
cmake, msvc and xcode project files, but those can be done later.

sezero:
> OK, I have nothing further.
> 
> The source files and any necessary defines & etc will still need adding to cmake, msvc and xcode project files, but those can be done later.

@slouken: Further review and final decision is for you.

connorjclark:
New conflict now... But good for merging otherwise?

Wohlstand:
Speaking about security, some year ago I watched Michael Pyne made a lot of tweaks and updates to resolve such problems, and I also helped him to fix the HES support that was broken accidentally. The topic you linked was made in 2016 and marked as resolved.

connorjclark:
@smcv As I understand it, for every subsystem of SDL_mixer the code is only compiled if requested. So by default, this wouldn't be introducing any security vulnerabilities. In other words, it's to the discretion of the library user, and if they want GME at all they are already fine with the risk.

connorjclark:
@slouken branch updated.

src/music.c
```diff
@@ -1259,6 +1295,32 @@ int Mix_PausedMusic(void)
...
+
+int Mix_GetNumTracks(Mix_Music *music)
+{
+    int result;
+
+    Mix_LockAudio();
+    if (music && music->interface->GetNumTracks) {
+        result = music->interface->GetNumTracks(music->context);
+    } else {
+        result = -1;
```
slouken, line 1318: `result = Mix_SetError("That operation is not supported");`

src/music.c
```diff
@@ -1259,6 +1295,32 @@ int Mix_PausedMusic(void)
     return (music_active == SDL_FALSE);
 }
 
+void Mix_StartTrack(int track)
+{
+    Mix_LockAudio();
+    if (music_playing && music_playing->interface->StartTrack) {
```
slouken, line 1301: This should return an int result code, and have an else clause with:
`result = Mix_SetError("That operation is not supported");`

include/SDL_mixer.h
```diff
@@ -2408,6 +2409,10 @@ extern DECLSPEC int SDLCALL Mix_PausedMusic(void);
  */
 extern DECLSPEC int SDLCALL Mix_ModMusicJumpToOrder(int order);
 
+/* Tracks */
+extern DECLSPEC void SDLCALL Mix_StartTrack(int track);
```
slouken, line 2413: How does this interact with starting and stopping the music playback? Does this start playing or does this just select the track to play? Should this maybe be `Mix_SelectTrack()` to more clearly show what is happening?

connorjclark: Currently it only changes the track. If the music isn't playing already, it doesn't do anything. SelectTrack sounds like a better name.

connorjclark: * Nothing happens if `Play` is not called before
* Otherwise, sets and plays the track. I think GME will restart a track if already playing

Not certain if point one is necessary. GME_StartTrack does call Play itself after all. Maybe the music_playing check should be removed in Mix_StartTrack

connorjclark: @Wohlstand any thoughts on the API here?

Wohlstand: I am fully against making new API calls without the `Mix_Music*` pointer. This is a very bad idea, and it makes a lot of problems in various cases, like, you are required to play the music until you will are able to change the stuff. Since I made MixerX, all new music-related API calls always require the music object pointer, which will allow you to manipulate music objects without the necessary to play them. I do remind some users who reported related bugs to my MixerX side that they are unable to retrieve meta-data until they start playing the music object.

Wohlstand: I rather make the:
```c
extern DECLSPEC void SDLCALL Mix_SelectTrack(Mix_Music *music, int track);
```
instead.

Wohlstand: Also, while this intended to select the track before to play, this is also a way that should allow switching the track on the fly without recreating the music object and reloading the file.

> P.S. Off topic: I think, I could use the same API at my MixerX side to allow switching sub-melodies of XMI files on the fly (if you didn't know, XMI may contain many different melodies inside, however, my sequencer doesn't support the proper switching between tracks yet, that I working on)

Wohlstand: P.S. Forgot to tell, at my MixerX you may also find several music-related API calls that repeating exist calls of SDL_Mixer, but they have a music object pointer. They were intended to manipulate individual music objects without starting to play them. In addition, they majorly helped me to implement multi-music API which allows playing multiple independent music streams and manipulating each music object independently. So, that is why I am totally against making new music-related calls without providing the music object pointer to it.

connorjclark: Agreed on the pointer thing.

But does GME's `gme_start_track` just switch the track or does it fail/do nothing if not already playing? Can't recall why I put a "Play" call in there.

Wohlstand: You may basically save the tracking number at the variable (at the internal structure) when music is inactive, and when starting its play, re-use the value from that variable. When music is running, also call the track switch (using that gme_start_track call).

connorjclark: Turns out gme does that for us.

connorjclark:
Updated!

src/codecs/music_gme.c
connorjclark: @sezero @Wohlstand

src/music.h
```diff
@@ -150,6 +151,11 @@ typedef struct
     /* Unload the library */
     void (*Unload)(void);
 
+    /* Start a specific track */
+    int (*StartTrack)(void *music, int track);
+
+    /* Get number of tracks. Returns -1 if not applicable */
+    int (*GetNumTracks)(void *music);
```
Wohlstand, line 158: Forgot to say, but I would put that under GetMetaTag() (or above) to keep the order of calls by role.

connorjclark:
Build failure seems unrelated: `Could NOT find libxmp`

madebr:
I recently pushed libxmp related updates for the cmake script.
So perhaps you need to rebase on top of current master.

connorjclark:
Anything else I can do?

src/music.h
```diff
@@ -150,6 +154,8 @@ typedef struct
     /* Unload the library */
     void (*Unload)(void);
 
+    /* Start a specific track */
+    int (*StartTrack)(void *music, int track);
```
slouken, line 158: Can you move `StartTrack` up after `GetNumTracks`?

src/codecs/music_gme.c
slouken: @connorjclark, can you please fix this as well?

src/codecs/music_gme.c
slouken: This is a good idea, but requires an ABI change. If you want, go ahead and enter an issue for that and we'll add it for SDL3_mixer.

src/music.h
slouken: I think this is ready to merge after you make that change.

connorjclark: Done!

sezero:
music_wavpack.c:Mix_MusicInterface_WAVPACK is missing the NULL entries
for the new GetNumTracks and StartTrack members.

src/codecs/music_cmd.c
```diff
@@ -293,6 +293,8 @@ Mix_MusicInterface Mix_MusicInterface_CMD =
     NULL,   /* LoopEnd */
     NULL,   /* LoopLength */
     NULL,   /* GetMetaTag */
+    NULL,   /* GetNumTracks */
+    NULL    /* StartTrack */
```
sezero, line 297: A comma is missing after NULL here.

sezero:
@slouken: Is this good to go?

slouken:
> @slouken: Is this good to go?

Yep!

sezero:
@madebr: this will need cmake'ry from you. Note that gme is disabled by default in autotools, gme git repo is https://bitbucket.org/mpyne/game-music-emu.git and it requires C++11 to build.

madebr:
Will libsdl-org vendor it?

Wohlstand:
> Note that gme is disabled by default in autotools

It can be used if installed into the system as a shared library, isn't it a problem?

sezero:
> Will libsdl-org vendor it?

If we want it? And I guess we do, since we do that for all other libs, and it's also possible that we may apply patches: @slouken?

sezero:
> > Note that gme is disabled by default in autotools
> 
> It still can be used if installed into the system as a shared library, isn't it a problem?

I don't understand the question here? Are we talking about the same thing?
GME option is disabled by default, because it isn't a very common necessity.
It can be enabled at build time by `--enable-music-gme` configury switch.

madebr:
Vendoring would allow us to include it in a sdl_mixer android archive, if we ever publish it on a public maven server.

sezero:
> Vendoring would allow us to include it in a sdl_mixer android archive, if we ever publish it on a public maven server.

OK, I went ahead and imported that bitbucket repository as https://github.com/libsdl-org/game-music-emu
No SDL-specific branch there: To be created, tweaked, etc, by whoever.

PR #380, opened 2022-05-03
music_ogg: fix sdl_seek_func() return value (SDL-2.0 / main)

sezero:
SDL_RWseek(), in both SDL-1.2 and SDL-2.0, returns the current offset upon
success and not NOERROR (0), therefore it behaves like `lseek`, not `fseek`
which vorbis expects. The current version seems to have worked so far by
luck.

CC: @Wohlstand

### DIFF
--- a/src/codecs/music_ogg.c
+++ b/src/codecs/music_ogg.c
@@ -169,7 +169,7 @@ static size_t sdl_read_func(void *ptr, size_t size, size_t nmemb, void *datasour
 
 static int sdl_seek_func(void *datasource, ogg_int64_t offset, int whence)
 {
-    return (int)SDL_RWseek((SDL_RWops*)datasource, offset, whence);
+    return (SDL_RWseek((SDL_RWops*)datasource, offset, whence) < 0)? -1 : 0;
 }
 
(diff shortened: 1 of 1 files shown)


Wohlstand:
Interesting... Usually the seek doesn't fail until the out-of-range value has been requested, or disk failed for any reason (like, got been disconnected or unmounted for any unexpected error, etc.). Normally I never met any seek errors, but could try...

EDIT: Yes, I confirm this (what going at Vorbis itself):
```c
static int _seek_helper(OggVorbis_File *vf,ogg_int64_t offset){
  if(vf->datasource){
    /* only seek if the file position isn't already there */
    if(vf->offset != offset){
      if(!(vf->callbacks.seek_func)||
         (vf->callbacks.seek_func)(vf->datasource, offset, SEEK_SET) == -1) // <--------------------
        return OV_EREAD;
      vf->offset=offset;
      ogg_sync_reset(&vf->oy);
    }
  }else{
    /* shouldn't happen unless someone writes a broken callback */
    return OV_EFAULT;
  }
  return 0;
}
```

PR #381, opened 2022-05-03
music_xmp: use xmp_load_module_from_callbacks() if available

sezero:
This way, we avoid loading the whole module into memory.

CC: @Wohlstand

### DIFF
--- a/src/codecs/music_xmp.c
+++ b/src/codecs/music_xmp.c
@@ -32,7 +32,14 @@
 #endif
 
 /* libxmp >= 4.5.0 constified several funcs */
+/* and also added load using file callbacks */
 #if (XMP_VERCODE < 0x040500)
+struct xmp_callbacks {
+    unsigned long (*read_func)(void *, unsigned long, unsigned long, void *);
+    int           (*seek_func)(void *, long, int);
+    long          (*tell_func)(void *);
+    int           (*close_func)(void*);
+};
 #define LIBXMP_CONST
 #else
 #define LIBXMP_CONST const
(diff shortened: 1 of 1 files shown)


src/codecs/music_xmp.c
```diff
@@ -89,6 +97,13 @@ static int XMP_Load(void)
         FUNCTION_LOADER(xmp_stop_module, void(*)(xmp_context))
         FUNCTION_LOADER(xmp_release_module, void(*)(xmp_context))
         FUNCTION_LOADER(xmp_free_context, void(*)(xmp_context))
+#if defined(XMP_DYNAMIC)
+        libxmp.xmp_load_module_from_callbacks = (int (*)(xmp_context,void*,struct xmp_callbacks)) SDL_LoadFunction(libxmp.handle, "xmp_load_module_from_callbacks");
```
Wohlstand, line 101: Reasonable to copy my `FUNCTION_LOADER_OPTIONAL` macro I made at my side (used for tempo code only, but also can be used here too):
```c
#ifdef XMP_DYNAMIC
#define FUNCTION_LOADER(FUNC, SIG) \
    libxmp.FUNC = (SIG) SDL_LoadFunction(libxmp.handle, #FUNC); \
    if (libxmp.FUNC == NULL) { SDL_UnloadObject(libxmp.handle); return -1; }
#define FUNCTION_LOADER_OPTIONAL(FUNC, SIG) \
    libxmp.FUNC = (SIG) SDL_LoadFunction(libxmp.handle, #FUNC);
#else
#define FUNCTION_LOADER(FUNC, SIG) \
    libxmp.FUNC = FUNC;
#define FUNCTION_LOADER_OPTIONAL(FUNC, SIG) \
    libxmp.FUNC = FUNC;
#endif
```

Wohlstand: The next code can be simplified then:
```c
#if (XMP_VERCODE >= 0x040500) || defined(XMP_DYNAMIC)
        FUNCTION_LOADER_OPTIONAL(xmp_load_module_from_callbacks, int (*)(xmp_context,void*,struct xmp_callbacks));
#else
        libxmp.xmp_load_module_from_callbacks = NULL;
#endif
```

PR #383, opened 2022-05-05
Use minor version for feature releases

smcv:
Same as libsdl-org/SDL_image#222, libsdl-org/SDL_image#223 and libsdl-org/SDL_image#224, but for SDL_mixer.

* autogen.sh: Fail early if a command fails

* dist: Don't try to distribute Xcode-iOS
    
    Fixes: 160959d8 "SDL_mixer cleanly builds on Apple platforms"

* workflows: Re-run autogen.sh

* workflows: Run `make distcheck`
    
    This confirms that we can build an official source tarball distribution.

* build: Use minor version number for feature releases
    
    Similar to what has been done in SDL itself, this allows micro versions
    with an odd minor version to be used as prereleases for a stable
    release. It also allows micro versions with an even minor version to
    be used to fix regressions in their corresponding .0 release
    if that becomes necessary.

* cmake: On Unix, make the library ABI-compatible with the Autotools build
    
    This is not yet implemented for macOS, only for generic Unix (in practice
    meaning Linux and *BSD).


sezero:
While you are at it, can you simplify autogen.sh like SDL's one?
```diff
diff --git a/autogen.sh b/autogen.sh
index 6441776..7598fea 100755
--- a/autogen.sh
+++ b/autogen.sh
@@ -9,11 +9,11 @@ echo "This may take a while ..."
 cat acinclude/* >aclocal.m4
-found=false
-for autoconf in autoconf autoconf259 autoconf-2.59
-do if which $autoconf >/dev/null 2>&1; then $autoconf && found=true; break; fi
-done
-if test x$found = xfalse; then
-    echo "Couldn't find autoconf, aborting"
-    exit 1
+
+if test "$AUTOCONF"x = x; then
+  AUTOCONF=autoconf
 fi
 
+$AUTOCONF || exit 1
+rm aclocal.m4
+rm -rf autom4te.cache
+
 # Run configure for this platform
```

smcv:
> can you simplify autogen.sh like SDL's one?

It can be even simpler than that.

sezero:
> Does this work with the old Bourne shell and dash and other non-gnu /bin/sh variants?

Tested by changing `#!/bin/sh` to `#!/bin/dash` worked for me
```
$ rpm -q dash
dash-0.5.5.1-4.el6.i686
```
@smcv and others may know better

smcv:
`${foo:-bar}` is POSIX shell syntax,  so it will work in any POSIX-compliant shell, likely including ancient proprietary Unixes.

Autoconf-generated `configure` scripts use `${foo-bar}`, which is from the same family as `:-` (the difference is that `:-` uses the fallback if the variable is set-but-empty), so I think that's good evidence of it being so portable that nobody cares about platforms that don't have it.

slouken:
Okay, thanks!

PR #384, opened 2022-05-07
updated github actions.

sezero:
This updates CI to test-build on linux and mingw with proper codecs
present (cfake support is immature as it has been, and I didn't pay
any extra attention to it.) Whole thing was adapted from  SDL_image.

There is room for impromevent here, I guess. E.g.: the clang option
of the old version can possibly be transplanted to it maybe?

Please review.

CC:  @smcv


slouken:
This seems reasonable to me.

sezero:
OK, will merge when CI completes.

@smcv: If you have improvements for this, patches are most welcome.

PR #388, opened 2022-05-17
cmake: fixed the use of SDL_mixer with disabled BUILD_SHARED_LIBS

rmg-nik:
- Fixed using SDL_mixer as subproject with cmake's add_subdirectory with disabled BUILD_SHARED_LIBS
- Removed checks for ANDROID when detecting SDL2 or SDL2-static

From now we can use SDL_mixer with cmake in this way:
```
option(BUILD_SHARED_LIBS "Build the library as a shared library" OFF)
option(BUILD_SAMPLES "Build the sample program(s)" OFF)
option(SDL_STATIC_PIC "Static version of the library should be built with Position Independent Code" ON) #for x86_64 Android build

add_subdirectory(external/SDL)
add_subdirectory(external/SDL_mixer)
add_subdirectory(path_to_your_project)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ option(SUPPORT_MID_TIMIDITY "Support loading MIDI music via TiMidity" ON)
 
 option(BUILD_SHARED_LIBS "Enable shared library" ON)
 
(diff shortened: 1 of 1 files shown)


madebr:
Thanks, once my (future pr of the) cmake rewrite of SDL_mixer + other SDL libraries is finished, I'll be tagging you for review.

PR #390, opened 2022-05-17
Fixes of the master_volume behaviour

Wohlstand:
- Make the `master_volume` variable being atomic
- Fixed the master_volume behavior at the music (there a middle buffer should be allocated to safely apply the volume effect)

Bonus: whitespace clean-up at the mixer.c.

### DIFF
--- a/src/music.c
+++ b/src/music.c
@@ -56,6 +56,7 @@ char *music_cmd = NULL;
 static SDL_bool music_active = SDL_TRUE;
 static int music_volume = MIX_MAX_VOLUME;
 static Mix_Music * volatile music_playing = NULL;
+static Uint8 *mix_music_buffer = NULL;
 SDL_AudioSpec music_spec;
 
 struct _Mix_Music {
@@ -310,20 +311,21 @@ int music_pcm_getaudio(void *context, void *data, int bytes, int volume,
 /* Mixing function */
 void SDLCALL music_mixer(void *udata, Uint8 *stream, int len)
 {
+    int master_vol = Mix_MasterVolume(-1);
     (void)udata;
 
     while (music_playing && music_active && len > 0) {
         /* Handle fading */
         if (music_playing->fading != MIX_NO_FADING) {
             if (music_playing->fade_step++ < music_playing->fade_steps) {
-                int volume = Mix_MasterVolume(-1);
+                int volume;
                 int fade_step = music_playing->fade_step;
                 int fade_steps = music_playing->fade_steps;
 
                 if (music_playing->fading == MIX_FADING_OUT) {
-                    volume = (volume * (music_volume * (fade_steps-fade_step))) / (fade_steps * MIX_MAX_VOLUME);
+                    volume = (music_volume * (fade_steps-fade_step)) / fade_steps;
                 } else { /* Fading in */
-                    volume = (volume * (music_volume * fade_step)) / (fade_steps * MIX_MAX_VOLUME);
+                    volume = (music_volume * fade_step) / fade_steps;
                 }
                 music_internal_volume(volume);
             } else {
@@ -339,7 +341,14 @@ void SDLCALL music_mixer(void *udata, Uint8 *stream, int len)
         }
 
         if (music_playing->interface->GetAudio) {
-            int left = music_playing->interface->GetAudio(music_playing->context, stream, len);
+            int left;
+            if (master_vol >= MIX_MAX_VOLUME || !mix_music_buffer) {
+                left = music_playing->interface->GetAudio(music_playing->context, stream, len);
+            } else {
+                SDL_memset(mix_music_buffer, music_spec.silence, (size_t)len);
+                left = music_playing->interface->GetAudio(music_playing->context, mix_music_buffer, len);
+                SDL_MixAudioFormat(stream, mix_music_buffer, music_spec.format, len, master_vol);
+            }
             if (left != 0) {
                 /* Either an error or finished playing with data left */
                 music_playing->playing = SDL_FALSE;
@@ -831,6 +840,9 @@ static int music_internal_play(Mix_Music *music, int play_count, double position
     /* Note the music we're playing */
     if (music_playing) {
         music_internal_halt();
+    }
+    if (!mix_music_buffer) {
+        mix_music_buffer = SDL_calloc(1, music_spec.size);
     }
     music_playing = music;
     music_playing->playing = SDL_TRUE;
@@ -1349,6 +1361,12 @@ void close_music(void)
 void unload_music(void)
 {
     size_t i;
+
+    if (mix_music_buffer) {
+        SDL_free(mix_music_buffer);
+        mix_music_buffer = NULL;
+    }
+
     for (i = 0; i < SDL_arraysize(s_music_interfaces); ++i) {
         Mix_MusicInterface *interface = s_music_interfaces[i];
         if (!interface || !interface->loaded) {


src/music.c
```diff
@@ -56,6 +56,7 @@ char *music_cmd = NULL;
 static SDL_bool music_active = SDL_TRUE;
 static int music_volume = MIX_MAX_VOLUME;
 static Mix_Music * volatile music_playing = NULL;
+static Uint8 *mix_music_buffer = NULL;
```
sezero, line 59: Is this one safe?

I understand that one can have multiple music instances loaded by Mix_LoadMUS,
but there is only one `music_playing` pointing only to one. I want to be sure though.

sezero:
CC: @rmg-nik

src/music.c
Wohlstand: This music buffer is used directly at the playing side only and has never been used in other cases. `music_mixer()` call is only one user here that works in one thread. 

Trivia: On the MixerX side, I have a similar buffer, but I use it to mix multiple music streams playing in parallel. Single-music player I kept being the same, and the multi-music player I made as a separate addition that works in parallel.

src/music.c
```diff
@@ -339,7 +341,14 @@ void SDLCALL music_mixer(void *udata, Uint8 *stream, int len)
         }
 
         if (music_playing->interface->GetAudio) {
-            int left = music_playing->interface->GetAudio(music_playing->context, stream, len);
+            int left;
+            if (master_vol >= MIX_MAX_VOLUME || !mix_music_buffer) {
```
sezero, line 345: Can `master_volume` ever be greater than `MIX_MAX_VOLUME`?

Wohlstand: Not intended to be, but just in a case.

src/music.c
```diff
@@ -321,9 +323,9 @@ void SDLCALL music_mixer(void *udata, Uint8 *stream, int len)
                 int fade_steps = music_playing->fade_steps;
 
                 if (music_playing->fading == MIX_FADING_OUT) {
-                    volume = (volume * (music_volume * (fade_steps-fade_step))) / (fade_steps * MIX_MAX_VOLUME);
+                    volume = (music_volume * (fade_steps-fade_step)) / fade_steps;
```
sezero, line 326: The initialization of volume a few lines above, i.e. `int volume = Mix_MasterVolume(-1);`
has no point any more: the initialization needs to go away.

Wohlstand: Forgot to remove here (at my side I already did that), thanks for the otice

src/music.c
sezero: I don't like it though. That buffer can become a member of `struct _Mix_Music`
and as such become per-music, can it not?

sezero: @slouken: What's your overall opinion, and your opinion about using that secondary buffer for music + master volume?

Wohlstand: >  and as such become per-music, can it not?

The idea of this buffer is the ability to separately mix multiple playing music streams (which I did at the MixerX side where I use this buffer to take a chunk of playing music with all applied effects and fade, then safely mix it into the output stream). You may review the implementation here: https://github.com/WohlSoft/SDL-Mixer-X/blob/master/src/music.c#L760-L858

sezero: We don't have a feature of playing and mixing multiple music streams though

Wohlstand: > We don't have a feature of playing and mixing multiple music streams though


I do think to try to submit it to mainstream Mixer once upon. This has many ways to use it:
- You can play heavy atmosphere environment sounds that aren't suitable for chunks API because of long loading and big memory usage.
- Allows running multiple parallel tracks of the same song and mute/unmuting each of them by necessary.
- Allows cross-fade of music from one to another. (by current music API it's basically impossible).

And, to keep the backward compatibility, I made a new set of functions that keeps the original API work as-is.

Anyway, as was discussed a while ago, some music engines won't work on multi-music API (such as MikMod and Native MIDI) as they were able to play only one song at the moment. So, in an attempt to load the thing using the multi-music API, the error will be returned.

sezero:
@slouken: Please review this:
- AFAICS, the atomic part should be OK
- the rest however, I question their merit here.

sezero:
> @slouken: Please review this:
> 
>     * AFAICS, the atomic part should be OK
> 
>     * the rest however, I question their merit (and implementation) here.

In addition, for master_volume affecting music, @rmg-nik's suggestion at
https://github.com/libsdl-org/SDL_mixer/pull/389#issuecomment-1129167303 _might_ be cleaner once implemented?

slouken:
Isn't this:
> Or fix the function music_internal_volume and make volume * master_volume / MIX_MAX_VOLUME
how we currently adjust music volume? Wouldn't we want to just adjust the music internal volume by the master volume?

Certainly doing a post-mix would work, but that's additional processing that's not needed in the music case. Am I missing something?

Wohlstand:
There is a bad idea to use music_internal_volume() for this case as there are API that allows setting an individual music volume per every music copy individually (`Mix_VolumeMusicStream()` and `Mix_GetMusicVolume()`). Otherwise, for user needs, the extra volume value is needed to be stored to split the user-defined and computed values.

sezero:
What is `Mix_VolumeMusicStream`? We don't have any such thing here.

Wohlstand:
So:
- post-mix way is simpler, it doesn't hurt anything (except for the necessity to allocate one small buffer piece and do another process)
- `music_internal_volume()` way: needs to specify the user-defined music individual volume, and computed volume after all volume factors are calculated into the value that will be physically used by the music itself.

> What is Mix_VolumeMusicStream? We don't have any such thing here.

Checked just now, and right, I forgot that I didn't port that part yet. That call I made at my side as a part of the preparation for the multi-music API at my side, mainly to allow control of each music entry independently.

sezero:
In any case: I am applying my atomic patch right now. Please rebase and drop the whitespace changes as well so that the rest can be reviewed more easily.

> > What is Mix_VolumeMusicStream? We don't have any such thing here.
> 
> Checked just now, and right, I forgot that I didn't port that part yet.

PLEASE STAY IN MAINSTREAM.

@slouken: The rest is up to you.

Wohlstand:
> Please rebase and drop the whitespace changes as well so that the rest can be reviewed more easily.

Done. Whitespace changes have been moved to here: #391.

P.S. In the next time, you can use the GitHub's reviewer functionality to completely hide any whitespace changes for convenience of review:


P.P.S. Attempting to post a reply, and there is a glitch...

slouken:
It looks like the atomic part of the change is gone now?

Wohlstand:
> It looks like the atomic part of the change is gone now?

Atomic got posted directly by @sezero, and I rebased my music fix patch.

Wohlstand:
Anything I need to change? Or all is okay for now?

sezero:
As I had said before, I questioned the implementation / its merit here. The alternative
solution for master_volume affecting music, @rmg-nik's suggestion at
https://github.com/libsdl-org/SDL_mixer/pull/389#issuecomment-1129167303 might be cleaner once implemented?

I leave things to @slouken's care / decision.

slouken:
This implementation adds an additional mixing step and associated CPU usage. Since the master volume is intended to affect all audio, I think it makes sense to have it affect music_internal_volume.

However, most games allow changing music and sound effect volume independently, so I think I'll just document that this function only affects sound effects and leave it at that for now.

Wohlstand:
> However, most games allow changing music and sound effect volume independently, so I think I'll just document that this function only effects sound effects and leave it at that for now.

Wise choice!

Then, it's needed to remove messy code from the music fade code simply, otherwise, it will make a bug.

slouken:
> Then, it's needed to remove messy code from the music fade code simply, otherwise, it will make a bug.

Yep, that is done in https://github.com/libsdl-org/SDL_mixer/commit/c0f1cffbbf8f693ad74a1be51b44d6f7fed661ba

Wohlstand:
btw, maybe rename the call to make it be more intuitive and don't confuse that it affects only SFX, not music streams? It's fine to do before the release will be done.

slouken:
All the other music functions have "Music" in the name, I think this is fine as-is.

Wohlstand:
No problem! :+1:

PR #394, opened 2022-05-21
minimp3: use SDL equivalents of standard C functions

Wohlstand:
On some platforms, this thing won't get built, and therefore suggested to use SDL's equivalents that will work.


sezero:
No: The overrides should happen in music_minimp3.c as defines before the minimp3.h includes.

Wohlstand:
I tried some before, but it leads even worse build failure as I will override definitions of original stdlib heads pulled at deeper places.

sezero:
We use that method at other places (see e.g. SDL_image)

Your issues probably come from memset, memcpy, etc: undefine before redefine solves it.
Will push a patch shortly.

slouken:
Don't bother with this change, I'm in the process of switching to dr_mp3.

PR #398, opened 2022-05-23
music_drflac.c: Fixed the crash on looped file loading

Wohlstand:
Fixed the crash because of a null-pointer access attempt, and make the proper loading of meta-data (sample-rate can't be accessed if not handled via `DRFLAC_METADATA_BLOCK_TYPE_STREAMINFO` block separately).

### DIFF
--- a/src/codecs/music_drflac.c
+++ b/src/codecs/music_drflac.c
@@ -50,6 +50,7 @@
 typedef struct {
     struct mp3file_t file;
     drflac *dec;
+    unsigned sample_rate;
     int play_count;
     int freesrc;
     int volume;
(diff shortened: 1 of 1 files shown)


src/codecs/music_drflac.c
```diff
@@ -87,11 +88,15 @@ static void DRFLAC_MetaCB(void *context, drflac_metadata *metadata)
...
-        drflac_uint32 i;
+    if (metadata->type == DRFLAC_METADATA_BLOCK_TYPE_STREAMINFO) {
+        music->sample_rate = metadata->data.streaminfo.sampleRate;
+        music->channels = metadata->data.streaminfo.channels;
+    } else if (metadata->type == DRFLAC_METADATA_BLOCK_TYPE_VORBIS_COMMENT) {
+        drflac_uint32 i, rate;
         char *param, *argument, *value;
         SDL_bool is_loop_length = SDL_FALSE;
         const char *pRunningData = (const char *)metadata->data.vorbis_comment.pComments;
+        rate = music->sample_rate;
```
sezero, line 99: in an unlikely case (such as with a broken or crafted file) where we don't hit a streaminfo tag, rate still will be 0: adding a paranoid check for it here would beneficial, e.g. like `if (!music->sample_rate) return;` or whatever.

Wohlstand: Or do you mean when metadata order is different inside the file? (when stream info goes after Vorbis comments). However, at the specification of FLAC format:

```
STREAMINFO: This block has information about the whole stream, like sample rate, number of channels, total number of samples, etc. It MUST be present as the first metadata block in the stream. Other metadata blocks MAY follow, and ones that the decoder doesn't understand, it will skip.
```
this block is required to be first, so, there is absolute guarantee the sample rate will be non-zero. For the worst case if a random file gets zero sample rate, inside the `_Mix_ParseTime()` there is no division operation with the samplerate. And the `music->sample_rate` is not used by other cases (where the `music->dec->sampleRate` is used directly). Anyway, if the sample rate value is zero, there is an invalid file that must be rejected.

sezero: Hm, after a brief look at dr_flac.h, unless one uses drflac_open_relaxed, such a file will be rejected, so we should be good.

sezero:
Thanks.

PR #400, opened 2022-05-27
mixer.c: Don't affect the music fade by master volume

Wohlstand:
Made instead of #390. Instead of making the master volume feature work with music too, there is a fixed bug that distorts the behavior of the music fade when the master volume is not maximum.


Wohlstand:
Ouch... gonna update my branch and push this again...

sezero:
Hasn't this already been done? https://github.com/libsdl-org/SDL_mixer/commit/c0f1cffbbf8f693ad74a1be51b44d6f7fed661ba

Wohlstand:
Right, didn't seen that yet.

PR #405, opened 2022-06-05
Don't wait for the music to fade out in Mix_FreeMusic if it is inactive

n1000:
Avoid a infinite loop / hang if the user's program has the music paused when trying to free it. If the music isn't active, it will never finish fading out.


sezero:
Sounds reasonable to me.  @slouken?

slouken:
Yep, looks good to me, thanks!

Issue #411, opened 2022-06-17
Static linking of dependent program with CMake fails in Autotools build with --disable-foo-shared


smcv:
To reproduce:

* Have a version of libflac that installs a pkg-config file
* Build SDL_mixer the way we do in Debian:
    * Autotools build
    * `--enable-music-flac-libflac`
    * `--disable-music-flac-libflac-shared` so that we rely on traditional static or dynamic linking rather than `dlopen()`
* Try to build the project in `cmake/test/` with `-DTEST_STATIC=ON` (the default)

(The same problem probably exists for dependencies other than libflac, I haven't tried.)

Expected result: `main_shared` is linked statically, as far as is possible.

Actual result:

```
[3/4] Linking C executable main_static
FAILED: main_static 
: && /usr/bin/cc   CMakeFiles/main_static.dir/main.c.o -o main_static  /usr/lib/x86_64-linux-gnu/libSDL2.a  /usr/lib/x86_64-linux-gnu/libSDL2_mixer.a  -lm -ldl  -lasound -lm -ldl -lpthread -lpulse-simple -pthread -lpulse -pthread  -lX11 -lXext -lXcursor -lXinerama -lXi -lXfixes -lXrandr -lXss -lXxf86vm -ldrm -lgbm -lwayland-egl -lwayland-client -lwayland-cursor -lxkbcommon -ldecor-0 -lpthread -lrt && :
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libSDL2_mixer.a(music_flac.o): undefined reference to symbol 'FLAC__stream_decoder_get_total_samples'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libFLAC.so.8: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
[4/4] Linking C executable main_shared
```

I think this is because `configure.ac` has two code paths for mandatory dependencies:

* if found via pkg-config: add to `PC_REQUIRES`, which ends up in `SDL2_mixer.pc` `Requires.private` but is not mentioned in `sdl2_mixer-config.cmake`
* else: add to `PC_LIBS`, which ends up in `SDL2_mixer.pc` `Libs.private` and in `sdl2_mixer-config.cmake` `_sdl2mixer_extra_static_libraries`

[...]

PR #412, opened 2022-06-17
Fix static linking to Autotools builds using CMake

smcv:
* autotools: Move mandatory dependencies from EXTRA_LDFLAGS to MIXER_LIBS
    
    Autotools variables whose names end with _LDFLAGS are usually for
    linker options such as -Wl,--export-dynamic which would not be
    appropriate to copy into the pkg-config or CMake metadata.
    
    Move these dependencies into a separate _LIBS variable, which is
    conventionally a list of -L and -l options (only) and *can* be copied
    into pkg-config or CMake metadata. This follows the same pattern as
    IMG_LIBS and TTF_LIBS in SDL_image and SDL_ttf.

    Exclude SDL_LIBS from this, because the CMake metadata has a special
    case for finding SDL, and linking to a different SDL causes link failure
    on our macOS CI. Link SDL via $(SDL_LIBS) instead.

* sdl2_mixer-config.cmake.in: Use MIXER_LIBS for static linking
    
    We can't rely on all mandatory dependencies being in PC_LIBS, because
    if we found a dependency such as libflac via pkg-config, we would have
    put the name of the pkg-config module in PC_REQUIRES instead.
    
    Resolves: https://github.com/libsdl-org/SDL_mixer/issues/411

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -25,6 +25,7 @@ CFLAGS  = @BUILD_CFLAGS@
 EXTRA_CFLAGS = @EXTRA_CFLAGS@
 LDFLAGS = @BUILD_LDFLAGS@
 EXTRA_LDFLAGS = @EXTRA_LDFLAGS@
+MIXER_LIBS = @MIXER_LIBS@
 LIBTOOL = @LIBTOOL@
 INSTALL = @INSTALL@
 AR	= @AR@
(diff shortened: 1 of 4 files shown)


madebr:
Thanks!
The cmake config script now works with all dependencies being non-dynamic.

Issue #421, opened 2022-07-12
2.5.2/2.6.0 regression: version number in SDL2_mixer.pc is empty


smcv:
To reproduce: Build SDL_mixer with Autotools, like Debian does. Install it. Run `pkg-config --modversion SDL2_mixer`

Expected result: `2.6.0` or whatever the version is.

Actual result: empty output

This causes dependent packages to fail to build if they have a pkg-config check for `SDL2_mixer >= 2.0.0` or similar.

---

Pull request will follow. I think we should cherry-pick the fix to the `release-2.6.X` branch and do a SDL_mixer 2.6.1 bugfix release, perhaps with a cherry-pick of b60b64f as well.

sezero:
> Build SDL_mixer with Autotools, like Debian does

@smcv: Briefly looked at your debian/rules file, here are a few minor notes:

- `--disable-music-mp3-mad-gpl` is no longer recognized: libmad support
  is removed.

- `--disable-music-mod-mikmod` is no longer recognized: mikmod support
  is removed.  Maybe you want `--disable-music-mod-xmp` instead (Although
  libxmp, especially v4.5.0 and newer should be preferred over libmodplug,
  IMO, but whatever.)

PR #422, opened 2022-07-12
configure.ac: Make sure to set VERSION

smcv:
Plain Autoconf only defines PACKAGE_NAME, PACKAGE_VERSION and similar
substitution variables, to which Automake adds PACKAGE and VERSION.
SDL_mixer doesn't use Automake, so it doesn't have the VERSION required
by our SDL2_mixer.pc.in (and SDL2_mixer.spec.in) unless we set it
explicitly.

Resolves: https://github.com/libsdl-org/SDL_mixer/issues/421  
Fixes: 52d3f2c "Rewrite CMake build script"

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,7 +31,10 @@ AC_SUBST(MINOR_VERSION)
 AC_SUBST(MICRO_VERSION)
 AC_SUBST(INTERFACE_AGE)
 AC_SUBST(BINARY_AGE)
-AC_SUBST(VERSION)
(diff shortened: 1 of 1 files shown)


smcv:
cc @madebr @slouken 

SDL_image and SDL_ttf do not have this bug, because they use the complete Autotools suite (Autoconf/Automake/Libtool) rather than just Autoconf.

SDL itself does not have this bug either, because although it only uses Autoconf, it substitutes a package-specific `@SDL_VERSION@` rather than relying on Automake's `@VERSION@`.

PR #424, opened 2022-07-13
Added Mix_PauseAudio() call to suspend/resume entire audio output

Wohlstand:
This call allows cheaper pausing/resuming of the whole audio processing without pausing every playing channel or music stream.

For example, this call can be used to pause the entire audio playback when the game is not focused.

Note: the "pause_async_music" call is required for music which is playing asynchronously from the main audio output processor, for example, OS Native MIDI. ~~However, the stuff doesn't affect CMD Music which will still play (however, never tested this as I never used CMD music for my needs)~~ Tested this, and it works, because it uses SIGSTOP and SIGCONT signals were literally suspended/resumes the running process.

### DIFF
--- a/include/SDL_mixer.h
+++ b/include/SDL_mixer.h
@@ -458,6 +458,15 @@ extern DECLSPEC int SDLCALL Mix_OpenAudio(int frequency, Uint16 format, int chan
 extern DECLSPEC int SDLCALL Mix_OpenAudioDevice(int frequency, Uint16 format, int channels, int chunksize, const char* device, int allowed_changes);
 
 /**
+ * Suspend or resume the whole audio output.
+ *
+ * \since This function is available since SDL_mixer 2.8.0.
+ *
+ * @param pause_on Set audio output to pause, 1 (to pause) or 0 (to resume)
+ */
+extern DECLSPEC void SDLCALL Mix_PauseAudio(int pause_on);
+
+/**
  * Find out what the actual audio device parameters are.
  *
  * If Mix_OpenAudioDevice() was called with `allowed_changes` set to anything
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -533,6 +533,15 @@ int Mix_OpenAudio(int frequency, Uint16 format, int nchannels, int chunksize)
                                 SDL_AUDIO_ALLOW_CHANNELS_CHANGE);
 }
 
+/* Pause or resume the audio streaming */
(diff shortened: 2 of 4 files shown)


sezero:
What is so special about the `pause_async_music` thing, and why not
simply use the existing Mix_PauseMusic and Mix_ResumeMusic instead
of it? Unless I'm missing something, you can do something like:
```c
void Mix_PauseAudio(int pause_on)
{
    SDL_PauseAudioDevice(audio_device, pause_on);
    if (pause_on) Mix_PauseMusic();
    else Mix_ResumeMusic();
}
```
And your `pause_async_music` misses setting `music_active`, I guess
it was not intentional.

Wohlstand:
Okay, now should be fine. Again, the pause_async_music() call is required for pausing/resuming asynchronous music streams only without any internal environment changes as this kills the idea of the Mix_PauseAudio() goal to "freeze the time". All other normal codecs get paused/resumed with the audio stream without any interaction at all and there is totally no worry.

include/SDL_mixer.h
```diff
@@ -457,6 +457,15 @@ extern DECLSPEC int SDLCALL Mix_OpenAudio(int frequency, Uint16 format, int chan
  */
 extern DECLSPEC int SDLCALL Mix_OpenAudioDevice(int frequency, Uint16 format, int channels, int chunksize, const char* device, int allowed_changes);
 
+/**
+ * Suspend or resume the whole audio output.
+ *
+ * \since This function is available since MixerX 2.7.0.
```
sezero, line 463: Should be `2.8.0` which will be the next release version

include/SDL_mixer.h
```diff
@@ -457,6 +457,15 @@ extern DECLSPEC int SDLCALL Mix_OpenAudio(int frequency, Uint16 format, int chan
...
+ *
+ * @param pause_on Set audio output to pause
```
sezero, line 465: This is unintelligible (at least to me.) You need something like `1 (for pause) or 0 (for resume)`, or something better.

Wohlstand: Done.

Wohlstand:
Ping?

slouken:
This seems good, although you probably want to document that it pauses all SDL audio processing, not just SDL_mixer.

Wohlstand:
Ye, this call is intended to pause entire SDL Audio processing.

PR #429, opened 2022-07-19
Fix FLAC target to link

AMDmi3:


sezero:
@madebr: Is this OK?

madebr:
Completely correct.
Apparantly, we were linking to the `libFLAC.so`/`libFLAC.a`/`FLAC.lib` library instead of the library found through [`FindFLAC`](https://github.com/libsdl-org/SDL_mixer/blob/main/cmake/FindFLAC.cmake).
This works for most cases, but not when e.g. FLAC comes from e.g. a binary release extracted in a random location.

PR #431, opened 2022-07-27
Expand CMake line a bit in README.txt

### LINKED ISSUE #430: Outdated README.TXT and update help
After commit `52d3f2c2a870d6b707b95fbfad6c814a66eb485e` my builds started failing because of options changes and/or something else that got changed in the cmake refactor. 

I am using SDL_mixer repo as a submodule of my project and adding the project via cmake. I was previously in commit `1daa96a48204569fcd1a07a768cf2dd92bc596d9`

According to Readme.txt file, I had to do e.g. `option(SUPPORT_OGG "" ON)` to enable stuff, but now they seem to be enabled by default. The options were renamed to `SDL2MIXER_*` instead, which is not reflected in the readme. 
Even replacing those, my builds are still failing with `Could NOT find opusfile (missing: opusfile_LIBRARY opusfile_INCLUDE_PATH)` even if the submodule is checked out and the libraries seems to be there. Not entirely sure how to migrate past that version.

Thanks in advance!

madebr:
Fixes #430

@ricardobusta
Does this clear things up a bit?

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -22,7 +22,7 @@ This library is under the zlib license, see the file "LICENSE.txt" for details.
 
 Note:
 Support for software MIDI, MOD, and Opus are not included by default because of the size of the decode libraries, but you can get them by running external/download.sh
-- When building with CMake, you can enable the appropriate SUPPORT_* options defined in CMakeLists.txt.
+- When building with CMake, you can enable the appropriate SDL2MIXER_* options defined in CMakeLists.txt. SDL2MIXER_VENDORED allows switching between system and vendored libraries.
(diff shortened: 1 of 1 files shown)


ricardobusta:
> Fixes #430
> 
> @ricardobusta Does this clear things up a bit?

Yup, thanks for updating it!

madebr:
Thanks for telling us the readme was outdated :)

slouken:
Do we need a similar update for SDL_image?

madebr:
Yes, I just pushed https://github.com/libsdl-org/SDL_image/commit/516cce263d254ed5aee7411e2969f63bb5ff5982

PR #437, opened 2022-08-08
Fix issue finding libxmp dependency

fjtrujy:
## Description
I was trying to compile `SDL_mixer` for `PS2` using the `cmake` functionality and it was always failing to try to find `libxmp`, however, it was properly installed.
After the change is done in this PR the library is properly detected.

Cheers

### DIFF
--- a/cmake/Findlibxmp-lite.cmake
+++ b/cmake/Findlibxmp-lite.cmake
@@ -1,3 +1,5 @@
+include(FindPackageHandleStandardArgs)
+
 find_library(libxmp_lite_LIBRARY
     NAMES xmp
 )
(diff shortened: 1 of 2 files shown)


fjtrujy:
> Thanks for finding and fixing this oversight, can you also add it to [`cmake/Findlibxmp-lite.cmake`](https://github.com/libsdl-org/SDL_mixer/blob/69e4e3744f9e29c3e643027623b43dad8d39fc51/cmake/Findlibxmp-lite.cmake)?

Yes, for sure!

fjtrujy:
Updated!

madebr:
And merged! Thanks again for the fix!

PR #447, opened 2022-09-19
Add sharedness-check to target_get_dynamic_library

### LINKED ISSUE #446: SDL_mixer misdetecting static library as dynamic library during build
I'm trying to set up Flatpak packaging for PySolFC, which depends on SDL_mixer via PyGame, and I've run into an odd misbehaviour in the dependency detection:

```
-- Found opusfile: /usr/lib/x86_64-linux-gnu/libopusfile.so  
-- Dynamic opus (opusfile): libopusfile.so.0
-- Enabled ogg music: using stb_vorbis
-- Using system libflac
-- Found FLAC: /usr/lib/x86_64-linux-gnu/libFLAC.so  
-- Dynamic libflac: libFLAC.so.8
-- Using system libmodplug
-- Found modplug: /app/lib/libmodplug.a  
-- Dynamic modplug: libmodplug.a
```

Note those last three lines. They result in this error when running the minimal reproducer build using `playmus.c`:

```
INFO: Opened audio at 44100 Hz 16 bit stereo 4096 bytes audio buffer
INFO: Couldn't load /app/bin/Astral_Dreams.it: Failed loading libmodplug.a: libmodplug.a: cannot open shared object file: No such file or directory
```

...which shows two issues:

1. As was foreshadowed, it's trying to load `libmodplug.a` as if it were a `.so` file.
2. For some reason, it's not picking up the `/app/lib` directory at runtime.

Omitting `*.a` from Flatpak's `cleanup` list and manually setting `LD_LIBRARY_PATH` to make it aware of `/app/lib` results in the following error, as expected:

```
[📦 com.github.libsdl_org.SDL_mixer.PlayMus ~]$ LD_LIBRARY_PATH=/app/lib playmus /app/bin/Astral_Dreams.it 
INFO: Opened audio at 44100 Hz 16 bit stereo 4096 bytes audio buffer
INFO: Couldn't load /app/bin/Astral_Dreams.it: Failed loading libmodplug.a: /app/lib/libmodplug.a: invalid ELF header
```

Now, it's entirely possible that I mis-built libmodplug, given that I just wrote this flatpak-builder source definition to avoid the binary "vendor all/none" option (I'm just trying to replace the SDL_mixer that comes with the Flatpak Runtime with a version that has libmodplug and FluidSynth support enabled, so I'd prefer to depend on the system libraries for everything else) and let it do the rest...

```json
    {
      "name": "libmodplug_sdl_vendored",
      "buildsystem": "cmake",
      "sources": [
        {
          "type": "git",
          "url": "https://github.com/libsdl-org/libmodplug",
          "commit": "f3173c24dcce28d54429099c555665a59d729546"
        }
      ]
    },

```

...but I'd call it a bug that SDL_mixer is picking up the `.a` file as a `.so` file at build time and producing a fundamentally broken runtime artifact.

madebr:
Fixes #446 by emitting an error


cmake/PrivateSdlFunctions.cmake
```diff
@@ -205,7 +212,30 @@ function(target_get_dynamic_library DEST TARGET)
             endif()
         endforeach()
     endif()
-    if (NOT result)
+    if(result)
+        string(TOLOWER "${result}" result_lower)
+        if(WIN32)
```
sezero, line 217: `if(WIN32 OR OS2)` , maybe?

madebr: Indeed. I added this to the patch.

sezero:
This is in, thanks. SDL itself and other satellite projects with the same functionality will need the same changes.

PR #450, opened 2022-09-20
switch libxmp to latest revision

sezero:
@slouken: make sure that this is OK for android, xcode and msvc builds.


sezero:
PING ?

slouken:
Still on my TODO list. I think I can check it today.

slouken:
Xcode and Android both work. VisualC isn't used to build libxmp here, so it should be fine.

slouken (approved):

sezero:
OK then, changeset pulled in.

PR #454, opened 2022-09-26
cmake: make FindVorbis.cmake compatible with upstream's VorbisConfig.cmake

### LINKED ISSUE #451: SDL2_mixer cmake configuration uses custom Findvorbisfile.cmake instead of system libvorbis files
SDL2_mixer includes and uses a Findvorbisfile.cmake to link with vorbisfile.
It also installs and uses said cmake module for use in the installed cmake configuration files for SDL2_mixer.

At least when building libvorbis with cmake myself, it installs its own configuration files to /lib/cmake/Vorbis, which exports Vorbis::vorbisfile and also handles necessary dependencies correctly, which is especially important for static builds.

The included Findvorbisfile.cmake on the other hand fails to correctly link with vorbisfile’s dependencies.

madebr:
Fixes #451


sezero:
You need to modify the install target in CMakeLists.txt (see CI failure)
```diff
-                cmake/Findvorbisfile.cmake
+                cmake/FindVorbis.cmake
```

PR #455, opened 2022-09-26
cmake: use target_link_libraries to add link options

### LINKED ISSUE #453: CMake: Native MIDI for macOS: frameworks are added as link_options instead of as link_libraries
With CMake under macOS, the frameworks needed for native MIDI are added as link_options instead of linked with target_link_libraries here: https://github.com/libsdl-org/SDL_mixer/blob/main/CMakeLists.txt#L728

This leads to the issue, that the frameworks are not exported into the CMake library configuration files, which leads to undefined references when linking to a static build of SDL2_mixer.

In SDL2, frameworks are linked with target_link_libraries, which seems to fix this problem.
See for example here:

- https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L2148
- https://github.com/libsdl-org/SDL/blob/main/CMakeLists.txt#L3114

madebr:
target_link_options cannot be used because it does not support static libraries

Fixes #453


sezero:
Does SDL_ttf need anything similar, btw?

madebr:
> Does SDL_ttf need anything similar, btw?

No. It does not do `target_link_options` on the `SDL2_ttf` target.

Issue #457, opened 2022-10-08
Android build with SDL2MIXER_FLAC=ON can't find libFLAC.so


genpfault:
The build can't find `libFLAC.so` with `SDL2MIXER_FLAC=ON` on an Android build, causing the build to fail:

```none
$ ./gradlew build

...

> Task :buildCMakeDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':buildCMakeDebug'.
> Expected output file at /project-name/android/build/intermediates/cxx/Debug/6w2m3w14/obj/arm64-v8a/libFLAC.so for target FLAC but there was none

...
```

`CMakeLists.txt`:

```cmake
cmake_minimum_required( VERSION 3.18.0 )
cmake_policy( SET CMP0091 NEW )
project( appname )

set( SDL_LIBC ON CACHE BOOL "" FORCE )
set( SDL_TEST OFF CACHE BOOL "" FORCE )
# https://github.com/libsdl-org/SDL.git
# release-2.24.1
add_subdirectory( external/sdl )

set( BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE )
set( SDL2MIXER_FLAC ON CACHE BOOL "" FORCE )
set( SDL2MIXER_OPUS OFF CACHE BOOL "" FORCE )
set( SDL2MIXER_SAMPLES OFF CACHE BOOL "" FORCE )
set( SDL2MIXER_VENDORED ON CACHE BOOL "" FORCE )
# https://github.com/libsdl-org/SDL_mixer.git
# HEAD of 'main' branch as of today: db9da3d416077195a2f84403b2fb40955518b4ab
add_subdirectory( external/sdl-mixer )

set( SRCS "src/main.cpp" )

set( LIBS
    SDL2::SDL2main
    SDL2::SDL2-static
    SDL2_mixer::SDL2_mixer-static
    )

if( ${CMAKE_SYSTEM_NAME} MATCHES "Android")
    add_library( main SHARED ${SRCS} )
    target_link_libraries( main ${LIBS} )
else()
    add_executable( ${CMAKE_PROJECT_NAME} ${SRCS} )
    target_link_libraries( ${CMAKE_PROJECT_NAME} ${LIBS} )
endif()
```

`src/main.cpp`:

```c++
#include <SDL.h>
#include <SDL_mixer.h>

int main( int argc, char** argv )
{
    SDL_Init( SDL_INIT_EVERYTHING );
[...]

PR #461, opened 2022-10-09
cmake: by default, disable building FLAC++

madebr:
While triaging https://github.com/libsdl-org/SDL_mixer/issues/457, I noticed the generated `app-debug.apk` contained `libFLAC++.so`.
Disabling building this library is the easiest way to avoid this.


madebr:
Looks like other dependency set `BUILD_TESTING`, will fix this shortly.

PR #470, opened 2022-11-20
Add missing error handling to SDL_realloc call

maxmitti:
Sorry, I previously forgot to error check the return value of SDL_realloc.
This is a minimal way of error handling, since I am not sure how realloc errors are appropriately handled in the code base.

So, feel free to edit the PR branch or let me know how I should change it.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -758,6 +758,7 @@ Mix_Chunk *Mix_LoadWAV_RW(SDL_RWops *src, int freesrc)
     SDL_AudioCVT wavecvt;
     int samplesize;
     int wavfree;        /* to decide how to free chunk->abuf. */
+    Uint8 *resized_buf;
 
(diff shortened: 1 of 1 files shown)


src/mixer.c
```diff
@@ -860,7 +860,12 @@ Mix_Chunk *Mix_LoadWAV_RW(SDL_RWops *src, int freesrc)
             return(NULL);
         }
 
-        chunk->abuf = SDL_realloc(wavecvt.buf, wavecvt.len_cvt);
+        Uint8 *const resized_buf = SDL_realloc(wavecvt.buf, wavecvt.len_cvt);
```
slouken, line 863: This won't compile on many systems, the declaration needs to be moved to the top of the block
`const Uint8 *resized_buf;`

maxmitti: Fixed.

PR #471, opened 2022-11-23
Fix samples install path

cgohlke:


madebr (approved):
Thank you for fixing this obvious typo!

PR #473, opened 2022-12-13
music_wavpack: do not setup buffer in S32 format for 8/16 bit sources.

sezero:
Avoids SDL's resampler e.g. for 16 bit files with 16 bit output setup,
which probably would be the most common case.  Wavpack library returns
the samples in 8, 16, 24, or 32 bit depths, but always in an int32_t[]
buffer in host-endian format; we now move the 8 bit and 16 bit samples
properly before feeding to SDL_AudioStream.

@icculus, @slouken: Please review for whether this would be beneficial
and if not, I will discard this.

P.S.: In case this accepted: Does it make any sense to further convert
8 bit samples from S8 to U8 format?

### DIFF
--- a/src/codecs/music_wavpack.c
+++ b/src/codecs/music_wavpack.c
@@ -194,7 +194,7 @@ typedef struct {
     void *decimation_cnxt;
 
     SDL_AudioStream *stream;
-    int32_t *buffer;
+    void *buffer;
     int32_t frames;
 
     Mix_MusicMetaTags tags;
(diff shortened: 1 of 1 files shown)


slouken (approved):
This seems like a good change.

slouken:
> P.S.: In case this accepted: Does it make any sense to further convert 8 bit samples from S8 to U8 format?

Yep, that's probably a good change as well. How common are signed 8-bit formats in wavpack?

sezero:
> This seems like a good change.

OK, will be merging shortly.

> > P.S.: In case this accepted: Does it make any sense to further convert 8 bit samples from S8 to U8 format?
> 
> Yep, that's probably a good change as well.

OK, will apply a patch after merging this one.

> How common are signed 8-bit formats in wavpack?

The answer is 'how common are signed 8-bit format wav files', and they
are (next to) none. The thing about S8 format used here is, the library
always returns the samples in signed format.

sezero:
Merged.

> > > P.S.: In case this accepted: Does it make any sense to further convert 8 bit samples from S8 to U8 format?
> > 
> > Yep, that's probably a good change as well.
> 
> OK, will apply a patch after merging this one.

Pushed https://github.com/libsdl-org/SDL_mixer/commit/e9b840794e9ee3cc9fe399ec55fdac6199da70b5 for it.

PR #477, opened 2022-12-21
mp3utils.c fixes

### LINKED ISSUE #476: Tag size seen as larger than file size error
I'm one of the pygame contributors, and we use SDL_mixer for all the audio decoding.

One of our users came in with this message printed to stdout every time they loaded a certain file:
`INFO: parse_id3v2: Tag size bigger than actual file size`

They went into a hex editor to investigate and found that this was not the case.

**The tag size is ox00000036 AKA 54**
ID3_tag_size_100wav

**After the header there are in fact 54 bytes**
ID3_tag_size_met_100wav

According to [the spec](https://id3.org/id3v2.3.0)
> The ID3v2 tag size is the size of the complete tag after unsychronisation, including padding, excluding the header but **not excluding the extended header** (total tag size - 10). Only 28 bits (representing up to 256MB) are used in the size description to avoid the introducuction of 'false syncsignals'.

The tag size should include the size of the extended header, I believe this is where the parser in SDL_mixer messes up.
https://github.com/libsdl-org/SDL_mixer/blob/main/src/codecs/mp3utils.c#L487-L507

If it sees an extended header it skips over it. But the tag length (`tag_len`) includes the size of the extended header, so the parser thinks its going to run over the end.

Here's the file (zipped into a folder, GitHub wouldn't let me just drop the WAV in):
[100.zip](https://github.com/libsdl-org/SDL_mixer/files/10265809/100.zip)

Wohlstand:
Fixes #476

Here are two changes done:
- Fixed the tag length failure because of the miscomputing at the extended header.
- Added more read error handling at `id3v22_parse_frame()` and `id3v2x_parse_frame()` functions.

### DIFF
--- a/src/codecs/mp3utils.c
+++ b/src/codecs/mp3utils.c
@@ -381,7 +381,7 @@ static size_t id3v22_parse_frame(Mix_MusicMetaTags *out_tags, struct mp3file_t *
     read_size = MP3_RWread(src, buffer, 1, ID3v2_2_FRAME_HEADER_SIZE);
 
     if (read_size < ID3v2_2_FRAME_HEADER_SIZE) {
-        SDL_Log("id3v2_parse_frame: Buffer size that left is too small %u < 6", (unsigned int)read_size);
+        SDL_Log("id3v22_parse_frame (1): Buffer size that left is too small (%u < 6)", (unsigned int)read_size);
         MP3_RWseek(src, frame_begin, RW_SEEK_SET);
         return 0; /* Buffer size that left is too small */
     }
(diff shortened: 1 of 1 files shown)


src/codecs/mp3utils.c
```diff
@@ -397,8 +397,18 @@ static size_t id3v22_parse_frame(Mix_MusicMetaTags *out_tags, struct mp3file_t *
 
     if (size < ID3v2_BUFFER_SIZE) {
         read_size = MP3_RWread(src, buffer, 1, size);
+        if (read_size < size) {
+            SDL_Log("id3v22_parse_frame (2): Buffer size that left is too small (%u < %u)", (unsigned int)read_size, (unsigned int)size);
```
Starbuck5, line 401: I don't understand these error changes, which is fine, but I do have a wording suggestion.

`Buffer size that left is too small` should be `Buffer size that is left is too small`. Or perhaps it would be better said as `Not enough buffer size remaining for read` or something.

Starbuck5 (approved):
This fixes the logged error statement! Thanks.

src/codecs/mp3utils.c
Wohlstand: This error is about the reading from the external file, and when we wanted to read, for example, 25 bytes, but had read 13 in fact. That means an error of an unexpected file end.

Wohlstand: I think I'll rephrase them into `unexpected end of the file` which is a true explanation of the error.

Wohlstand: Done.

PR #479, opened 2022-12-23
cmake: add gme support

madebr:
This adds cmake support for building sdl_mixer with GME music support.

todo:
- [x] libgme is currently unable to be built as a vendored library (like `modplug` and `fluidsynth`).


sezero:
Also needs applying to SDL3 version (with porting is necessary.)

madebr:
gme is added to SDL3_nixer in #483

PR #481, opened 2022-12-24
add Mix_SetMusicBeat and Mix_GetMusicBeat

connorjclark:
I have a use case where I must set the position of MIDI music to a particular beat index (something supported by the old allegro 4 midi player, and perhaps other midi players).

Example MIDI: https://drive.google.com/file/d/1qiE7sVyvQoVRxJftKhz-n4juuArfwri8/view?usp=share_link

And positions of tunes within:

```
0:    Wind Fish
81:   Overworld
233:  Hyrule Castle
553:  Lost Woods
814:  Great Sea
985:  East Hyrule
1153: Dancing Dragon
1333: Stone Tower
1556: Villages
1801: Swamp + Desert
2069: Outset Island
2189: Kakariko Village
2569: Clock Town
2753: Temple
2856: Dark World
3042: Dragon Roost
3125: Horse Race
3217: Credits
3296: Zelda's Lullaby
```

Note: I'm not certain if there's an off-by-one error in this code, or in the source midi player these positions are defined to work in (allegro 4), but I found I need to take the beat values above and subtract one to get the exact right positioning.

You can use the `playmus` example program to seek to a specific tune (you'll need a timidity.cfg and soundfont within your build directory, as always):

```
./playmus -n 552 TheTravelsOfLink.mid
```

Draft until some feedback (then I'll update the other music interfaces with NULL stubs).

### DIFF
--- a/src/codecs/timidity/playmidi.c
+++ b/src/codecs/timidity/playmidi.c
@@ -560,6 +560,75 @@ static void seek_forward(MidiSong *song, Sint32 until_time)
   song->current_sample=until_time;
 }
 
+static void seek_forward_beat(MidiSong *song, Uint32 beat)
+{
+  reset_voices(song);
+  while (song->current_event->beat < beat)
+    {
+      switch(song->current_event->type)
+	{
+	  /* All notes stay off. Just handle the parameter changes. */
+
+	case ME_PITCH_SENS:
+	  song->channel[song->current_event->channel].pitchsens =
+	    song->current_event->a;
+	  song->channel[song->current_event->channel].pitchfactor = 0;
+	  break;
+
+	case ME_PITCHWHEEL:
+	  song->channel[song->current_event->channel].pitchbend =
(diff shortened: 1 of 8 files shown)


src/codecs/timidity/playmidi.c
```diff
@@ -560,6 +560,75 @@ static void seek_forward(MidiSong *song, Sint32 until_time)
   song->current_sample=until_time;
 }
 
+static void seek_forward_beat(MidiSong *song, Uint32 beat)
+{
+  reset_voices(song);
+  while (song->current_event->beat < beat)
+    {
+      switch(song->current_event->type)
```
connorjclark, line 568: Exact same contents in this switch statement as the other seek_foward function. I could split out to a help function.

sezero:
@icculus seems to have removed library/format specific apis from SDL3_mixer. Closing this one now.

PR #483, opened 2023-01-06
CMake updates for SDL3

madebr:


Xcode/pkg-support/resources/CMake/sdl3_mixer-config.cmake
```diff
@@ -0,0 +1,66 @@
...
+set(SDL3_mixer_FOUND                TRUE)
+
+set(SDL3MIXER_VENDORED              TRUE)
+
+set(SDL3MIXER_CMD                   FALSE)
+
+set(SDL3MIXER_FLAC_LIBFLAC          FALSE)
+set(SDL3MIXER_FLAC_DRFLAC           TRUE)
+
+set(SDL3MIXER_GME                   FALSE)
```
madebr, line 22: I'm assuming here the SDL3_mixer framework won't ship with gme support.

slouken: I haven't looked at that yet, but I can flip that on if it does.

VisualC/pkg-support/cmake/sdl3_mixer-config.cmake
```diff
@@ -0,0 +1,78 @@
...
+set(SDL3_mixer_FOUND                TRUE)
+
+set(SDL3MIXER_VENDORED              TRUE)
+
+set(SDL3MIXER_CMD                   FALSE)
+
+set(SDL3MIXER_FLAC_LIBFLAC          FALSE)
+set(SDL3MIXER_FLAC_DRFLAC           TRUE)
+
+set(SDL3MIXER_GME                   FALSE)
```
madebr, line 21: Same here, assume the VC SDK does not have gme support.

slouken: No, it doesn't.

Xcode/pkg-support/resources/CMake/sdl3_mixer-config.cmake
sezero: SDL2 version doesn't enable it by default either due to GME being not a very common necessity

slouken:
Thanks! :)

PR #494, opened 2023-01-27
music_wavpack: disable DSD music by default,

sezero:
Use `--enable-music-wavpack-dsd` configury switch to enable it.

If this is accepted, @madebr should handle the cmake side of things.

### DIFF
--- a/Makefile.os2
+++ b/Makefile.os2
@@ -37,6 +37,8 @@ USE_MPG123=yes
 USE_DRMP3=no
 # wavpack music support
 USE_WAVPACK=yes
+# wavpack DSD music support
+USE_WAVPACK_DSD= no
 # midi music support (using timidity)
 USE_TIMIDITY=yes
 # midi music support (using fluidsynth)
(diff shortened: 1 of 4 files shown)


slouken:
This sounds fine, but what's the reasoning for this?

sezero:
The decimation code is not easy on the cpu, and DSD files aren't your everyday format to be used in games. And it _is_ configurable, so nothing is actually lost.

madebr:
Has SDL_mixer access to hints? This looks like something that can be done without a compile switch.

sezero:
> Has SDL_mixer access to hints? This looks like something that can be done without a compile switch.

It _should_ have access to hints, I guess. Just not something we ever had done in the satellite libs for something like this.

slouken:
Or maybe we can just leave the support in, and it'll only be used by people who load DSD music?

sezero:
> Or maybe we can just leave the support in, and it'll only be used by people who load DSD music?

That's surely OK too. Should I close this?

slouken:
Yes, go ahead.

sezero:
OK, closing.

PR #495, opened 2023-02-09
Fix potential infinite loop in `Mix_FreeChunk` when called in a channel finish callback

MCJack123:
In a recent update to my app's dependencies, I noticed that my channel finished callback, which simply frees the current chunk with `Mix_FreeChunk`, was causing a stack overflow through infinite recursion. This appears to stem from a change in SDL2_mixer 2.6.0 that moved some logic for halting channels around - the channel finished callback is now called *before* setting the channel as not playing, which means that calling any functions that may attempt to halt the channel, including `Mix_FreeChunk`, inside the callback will result in an infinite loop.

This PR adjusts `Mix_HaltChannel_locked` to call `_Mix_channel_done_playing` *after* setting the channel to not playing, which will prevent these sorts of loops. Note that I have yet to test this change, but it shouldn't result in anything major - other code in the file does it in this same order, calling `_Mix_channel_done_playing` only after setting the two fields to 0.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -926,9 +926,9 @@ Mix_Chunk *Mix_QuickLoad_RAW(Uint8 *mem, Uint32 len)
 static void  Mix_HaltChannel_locked(int which)
 {
     if (Mix_Playing(which)) {
-        _Mix_channel_done_playing(which);
         mix_channel[which].playing = 0;
         mix_channel[which].looping = 0;
+        _Mix_channel_done_playing(which);
     }
     mix_channel[which].expire = 0;
     if (mix_channel[which].fading != MIX_NO_FADING) /* Restore volume */


slouken:
I think this is okay, can you let me know if you find any issues in your testing?

sezero:
Looks like issue started with https://github.com/libsdl-org/SDL_mixer/commit/5d3c84e5dd4c91093255ad5533b8457ecc40684f: CC @pionere

If accepted, this should go into SDL2 branch too.

pionere:
> Looks like issue started with [5d3c84e](https://github.com/libsdl-org/SDL_mixer/commit/5d3c84e5dd4c91093255ad5533b8457ecc40684f): CC @pionere

Are you sure about that? The code from line 841..843 is the same as it was in line 1116..1118, which means the infinite loop existed before that if the Mix_HaltChannel was called for an explicit channel.

The change seems to be reasonable to me, though.

sezero:
> > Looks like issue started with [5d3c84e](https://github.com/libsdl-org/SDL_mixer/commit/5d3c84e5dd4c91093255ad5533b8457ecc40684f): CC @pionere
> 
> Are you sure about that?

Nope, not after re-reading that commit after your comment

> The change seems to be reasonable to me, though.

OK, thanks.

@slouken: Let's merge?

PR #502, opened 2023-03-02
Remove unneed 'volatile' keyword

1bsyl:
No volatile keyword need right ? 
if there is an issue, it should be done with mutex / atomic


1bsyl:
merging because there is no need of volatile keyword here

PR #504, opened 2023-03-08
CMake: Improve dependency searching

FtZPetruska:
## Description

This PR brings a set of changes that should help searching for dependencies. Feedback is greatly appreciated!

Here is a summary of the changes.

### 1. Matching the name with upstream

FluidSynth exports the target `FluidSynth::libfluidsynth`.

In the exported config, the casing of `find_dependency(MPG123)` should be `mpg123`, and `opusfile::opusfile` should be `OpusFile::opusfile`.

### 2. Add `@PACKAGE_INIT@` to the config template

While the config file was already configured using the `configure_package_config_file` helper, it did not have a `@PACKAGE_INIT@`.

When expanded, it provides the `set_and_check` and `check_required_components` macros, as well as setting `PACKAGE_PREFIX_DIR` to the correct directory.

### 3. Correct `INTERFACE_LINK_FLAGS`

This property [does not seem to exist](https://cmake.org/cmake/help/latest/search.html?q=INTERFACE_LINK_FLAGS). The `INTERFACE_LINK_OPTIONS` property seems to be what was wanted ([doc](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_OPTIONS.html)).

### 4. Get help from pkg-config when available

While a user can manually set for a dependency the `Foo_COMPILE_OPTIONS`, `Foo_LINK_LIBRARIES`, and `Foo_LINK_OPTIONS` to provide the usage requirements, we can use pkg-config when available to pre-fill these cache variables.

This works as follows:
- call `find_package(PkgConfig QUIET)`, we want the search to be quiet as this is just optional help
- search for the module `pkg_check_modules(PC_FOO QUIET foo)`, similarly we want the search to be quiet. Note that it is not necessary to guard this call with `PKG_CONFIG_FOUND`, if pkg-config is not found this call will simply be ignored.
- use the `get_flags_from_pkg_config("${Foo_LIBRARY}" "PC_FOO" "_foo")` helper to get the appropriate set of flags. We use the library suffix to infer whether we should take the static or dynamic flags and prefix them with `_foo`.
- Use these variables to initialise the `Foo_COMPILE_OPTIONS`, `Foo_LINK_LIBRARIES`, `Foo_LINK_OPTIONS`, and `Foo_LINK_DIRECTORIES` cache variables. If pkg-config is unavailable they will simply be empty as they currently are.

For this to work better, the policy [CMP0099](https://cmake.org/cmake/help/latest/policy/CMP0099.html) is set to `NEW`. This covers this very specific use case:
- Have libogg installed in a non-standard directory such as `/opt/my_libs`
- Build libflac statically with this custom libogg.
- Build SDL_mixer statically with examples.
- Build fails due to `ogg not found` 

While `FLAC::FLAC`'s `INTERFACE_LINK_DIRECTORIES` will be properly set `/opt/my_libs`, this will not be propagated unless `CMP0099` is set to `NEW`

### 5. Add logic when vcpkg or conan is used

When either vcpkg or conan is used:
- Turn off vendored libraries by default
- Set `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to `ON`

The latter option makes CMake prioritise using `*-config.cmake`/`*Config.cmake` over the find modules. Both package managers install the upstream CMake config when available, which will commonly include the usage requirements.

### 6. Add alternative names for libraries

Most notably on Windows, static builds of libraries may have a different name from their shared counterpart. For example, a static build of FluidSynth on Windows will produce `libfluidsynth.lib` which would not be found.

---

## Testing

The changes were tested on:
- macOS
  - Using shared libraries provided by Homebrew
  - Using static libraries provided by vcpkg
- Windows
  - Using the `x64-windows` triplet on vcpkg
  - Using the `x64-windows-static` triplet on vcpkg
- 3DS Hombrew
  Platform such as this one that only support static linking and which do not have vcpkg support benefit a lot from the help provided by pkg-config. Previously, making a build with libFLAC, tremor, and opusfile would be a dependency nightmare for linking, but with these changes, the linking process is now trivial. I expect a similar situation for PSP and PSVita Homebrew.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -928,6 +928,7 @@ if(SDL3MIXER_INSTALL)
         FILES
             "${CMAKE_CURRENT_BINARY_DIR}/SDL3_mixerConfig.cmake"
             "${CMAKE_CURRENT_BINARY_DIR}/SDL3_mixerConfigVersion.cmake"
+            "${CMAKE_CURRENT_LIST_DIR}/cmake/PkgConfigHelper.cmake"
             cmake/FindFLAC.cmake
             cmake/FindFluidSynth.cmake
             cmake/Findgme.cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,16 +47,11 @@ endif()
 
 # Assume MSVC projects don't have a package manager and need vendored dependencies (by default).
 # Most other platforms have some kind of package manager.
-if(ANDROID OR MSVC AND NOT VCPKG_TOOLCHAIN AND NOT CONAN_CMD)
+# FIXME: consider a package manager such as conan/vcpkg instead of vendoring
+if(ANDROID OR MSVC)
     set(vendored_default ON)
 else()
     set(vendored_default OFF)
-endif()
-
-# Both vcpkg and conan provide the upstream CMake config for the
-# dependencies, so prioritise them over the Find modules
-if(VCPKG_TOOLCHAIN OR CONAN_CMD)
-    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
 endif()
 
 set(sdl3mixer_install_enableable ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,11 +47,16 @@ endif()
 
 # Assume MSVC projects don't have a package manager and need vendored dependencies (by default).
 # Most other platforms have some kind of package manager.
-# FIXME: consider a package manager such as conan/vcpkg instead of vendoring
-if(ANDROID OR MSVC)
+if(ANDROID OR MSVC AND NOT VCPKG_TOOLCHAIN AND NOT CONAN_CMD)
     set(vendored_default ON)
 else()
     set(vendored_default OFF)
+endif()
+
+# Both vcpkg and conan provide the upstream CMake config for the
+# dependencies, so prioritise them over the Find modules
+if(VCPKG_TOOLCHAIN OR CONAN_CMD)
+    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
 endif()
 
 set(sdl3mixer_install_enableable ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,13 @@ if(POLICY CMP0112)
     cmake_policy(SET CMP0112 NEW)
 endif()
 
+if(POLICY CMP0099)
+    # Make `INTERFACE_LINK_DIRECTORIES` a transitive usage requirement.
+    # This is needed for static dependencies which have transitive dependencies
+    # outside of compiler default search paths.
+    cmake_policy(SET CMP0099 NEW)
+endif()
+
 # Assume MSVC projects don't have a package manager and need vendored dependencies (by default).
 # Most other platforms have some kind of package manager.
 # FIXME: consider a package manager such as conan/vcpkg instead of vendoring
@@ -58,6 +65,8 @@ include(CheckSymbolExists)
 include(CMakeDependentOption)
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
+
+include(PkgConfigHelper)
 
 option(CMAKE_POSITION_INDEPENDENT_CODE "Build static libraries with -fPIC" ON)
 option(BUILD_SHARED_LIBS "Build the library as a shared library" ON)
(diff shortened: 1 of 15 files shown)


madebr:
All patches look good to me, apart from 8da1b011ed6033d347bf5eef937bde5329de0887
I'd prefer to not have any packaging specific code in the cmake build script.
Instead, the vcpkg and conan-center-index build recipes should be updated to pass:
`-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON -DSDL3MIXER_VENDORED=OFF` to cmake.

What do you think?

FtZPetruska:
> Also, what do you think about the current `vendored_default` logic? Would it surprise people that the cmake script, by default vendors its dependencies on Android and MSVC, but not on other platforms?

I think that is a good default, when using MSVC or an Android toolchain, there is no "default" package manager to grab the dependencies from, usually the user will have to compile them from source. Whereas on both Linux and macOS, it is pretty much guaranteed to have a package manager that provides development libraries.

> All patches look good to me, apart from [8da1b01](https://github.com/libsdl-org/SDL_mixer/commit/8da1b011ed6033d347bf5eef937bde5329de0887) I'd prefer to not have any packaging specific code in the cmake build script. Instead, the vcpkg and conan-center-index build recipes should be updated to pass: `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON -DSDL3MIXER_VENDORED=OFF` to cmake.
> 
> What do you think?

While editing the upstream build scripts to include both options would work for the most part, users still need to manually specify `CMAKE_FIND_PACKAGE_PREFER_CONFIG` when:
- They use conan/vcpkg to only provide the dependencies, not using the upstream build scripts
- They consume the library afterwards (whether it was built from the build script or manually)

This commit was aiming to fix this comment specifically about considering vcpkg and conan to determine the defaults:
```cmake
# FIXME: consider a package manager such as conan/vcpkg instead of vendoring
```

But I can understand if you think that is not a satisfying solution. Let me know if you want me to revert the commit 👍

madebr:
I'm open to enabling `CMAKE_FIND_PACKAGE_PREFER_CONFIG` unconditionally, and treat our own `FindXXX.cmake` modules as backup.

madebr:
What do you think about the following:
- keep the vendored handling as it is now, and require the package maintainers to pass `-DSDL3MIXER_VENDORED=FALSE`
- set `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to true unconditionally in the cmake script.

FtZPetruska:
Sounds good to me, I will push the changes shortly 👍

CMakeLists.txt
```diff
@@ -47,18 +47,13 @@ endif()
 
 # Assume MSVC projects don't have a package manager and need vendored dependencies (by default).
 # Most other platforms have some kind of package manager.
-if(ANDROID OR MSVC AND NOT VCPKG_TOOLCHAIN AND NOT CONAN_CMD)
+# FIXME: consider a package manager such as conan/vcpkg instead of vendoring
```
madebr, line 50: I'm going to remove this comment, because it doesn't look like we'll use a package manager ourselves anytime soon.

madebr (approved):

sezero:
@madebr, @FtZPetruska: About opusfile and _matching upstream's target names_: Opusfile latest official release is 0.12 and it doesn't have cmake support -- our fork had copied cmake support from opusfile development at some point. NOW, current opusfile development has this change: https://github.com/libsdl-org/opusfile/commit/c7853ce10dd53eb26b32bfc001159c3f718faf2f. I'd like to upgrade our vendored opusfile to current opusfile master: How does it affect us here in SDL_mixer - any changes needed?

madebr:
> @madebr, @FtZPetruska: About opusfile and _matching upstream's target names_: Opusfile latest official release is 0.12 and it doesn't have cmake support -- our fork had copied cmake support from opusfile development at some point. NOW, current opusfile development has this change: [libsdl-org/opusfile@c7853ce](https://github.com/libsdl-org/opusfile/commit/c7853ce10dd53eb26b32bfc001159c3f718faf2f). I'd like to upgrade our vendored opusfile to current opusfile master: How does it affect us here in SDL_mixer - any changes needed?

Looks like none are needed. That commit you referred to is made for Windows <-> Linux case sensitivity compatibility.
We're doing `find_package(OpusFile)`, and the file `OpusFileConfig.cmake` remains existing, so we're fine.

PR #506, opened 2023-03-13
Update mixer.c to use SDL3 Atomic type

Th3T3chn0G1t:
Just fixes a build error introduced by the use of `SDL_atomic_t` in mixer.c instead of `SDL_AtomicInt`


1bsyl:
merged!

PR #517, opened 2023-05-05
replace Mix_* with MIX_*

### LINKED ISSUE #505: SDL Mixer API namespace consistency
The SDL and SDL image APIs capitalize the namespace prefix:

- SDL_Init
- IMG_Init

The SDL mixer and SDL net APIs do not:

- Mix_Init
- Net_Init

Unless this was deliberate, the transition to SDL3 would be a great time to make these APIs consistent.

Nhalrath:
Closes #505.

### DIFF
--- a/include/SDL3/SDL_mixer.h
+++ b/include/SDL3/SDL_mixer.h
@@ -100,7 +100,7 @@ extern "C" {
  *
  * \since This function is available since SDL_mixer 2.0.0.
  */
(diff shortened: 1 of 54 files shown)


sezero:
I don't like this: SDL_mixer has had its namespace for years.

CC: @icculus, @slouken

icculus:
Yeah, I agree. I don't love "Mix_" but I don't think we should change it at this point.

PR #521, opened 2023-05-12
Add seek, tell, and duration to fluidsynth

### LINKED ISSUE #519: Support Seek and other functions for fluidsynth 2+
Fluidsynth 2 supports:
Seek: https://www.fluidsynth.org/api/group__midi__player.html#gaab5fe5b25ff513919b138d77e079c75a
Tell: https://www.fluidsynth.org/api/group__midi__player.html#gabc2d114e2a6926f4db800286b2ebb1da
Duration: https://www.fluidsynth.org/api/group__midi__player.html#gab8bec2b53e07daf2313412e4663c0fa7

When Fluidsynth 2 is used, these functions should be available.

This would fix an issue for us where Mix_PlayMusic continues from where it left off instead of the beginning of the track: https://github.com/CorsixTH/CorsixTH/issues/2328

TheCycoONE:
Ensure that when playing a track it starts from the beginning.

Fixes #519

### DIFF
--- a/src/codecs/music_fluidsynth.c
+++ b/src/codecs/music_fluidsynth.c
@@ -39,6 +39,9 @@ typedef struct {
 #if (FLUIDSYNTH_VERSION_MAJOR >= 2)
     void (*delete_fluid_player)(fluid_player_t*);
     void (*delete_fluid_synth)(fluid_synth_t*);
+    int (*fluid_player_seek)(fluid_player_t*, int);
+    int (*fluid_player_get_total_ticks)(fluid_player_t*);
+    int (*fluid_player_get_current_tick)(fluid_player_t*);
 #else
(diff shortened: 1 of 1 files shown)


TheCycoONE:
The change cherry-picks cleanly onto main if you prefer. I am working with sdl2 so I targeted that branch.

madebr:
Please create a PR for sdl3 as well.
So we can keep both branches in sync.

slouken:
Merged and cherry-picked to main, thanks!

sezero:
The tick functions are actually available since v1.1.7: should we still restrict them to builds against v2.0?

TheCycoONE:
Both those releases were in 2018, I'm not sure it's worth the extra condition?

PR #527, opened 2023-05-31
[SDL3] Fixed detection of FLAC files with ID3 tag at begin

Wohlstand:
Recently I found that some FLAC files won't play at all. I had checked, and I found it begins with an ID3 tag. Also, I checked around the stuff, and it's not a paranomrmal mess of Audacity: it's one of legacy forms of FLAC files that used ID3 tags at begin.

There is a comment where I explained about this a first time: https://github.com/WohlSoft/SDL-Mixer-X/commit/94e87503c80cc0311dfedb370e21d25573492e33

Example file (I exported it in 2019'th year by the old version of Audacity): [Fox fantasy.mid-battle-chess-gmized.flac.zip](https://github.com/libsdl-org/SDL_mixer/files/11607303/Fox.fantasy.mid-battle-chess-gmized.flac.zip)


sezero:
Like I said in original commit: The idea of FLAC files with ID3 tags??
I shall vomit...

This may be one way of handling it. (Labeling a file starting with ID3
as mp3 is flaky to start with...) Any other ways to do this?

Wohlstand:
Maybe use my own implementation of `detect_mp3()`? https://github.com/WohlSoft/SDL-Mixer-X/blob/master/src/music.c#L1385-L1456

I mean, call it after skipping the ID3 tag (and inside the `detect_mp3()` itself, I should remove the check for the presense of `ID3`).

I polished it for many years on thousands of various MP3 files that I had at my personal collections.

Does this sounds better?

sezero:
> Like I said in original commit: The idea of FLAC files with ID3 tags?? I shall vomit...

And I wonder: how much should we care about flac files with id3?
I.e.: how common are they?

CC: @ktmf01 as libflac maintainer.

Wohlstand:
At the mainstream FLAC I also do see the code that skips ID3 tags: https://github.com/xiph/flac/blob/1619af5a36fc0343cdf6b3517bb78d8aee85fe59/src/libFLAC/metadata_iterators.c#L3143-L3176

And at the changelog I see, the ID3 tags were being supported in the past, but had been removed, and now, if any FLAC file contains these tags, they just gets skipped.

ktmf01:
> And at the changelog I see, the ID3 tags were being supported in the past

We're talking almost 19 years ago here. Support from plugins was removed in October 2004, and the command line programs never supported reading or writing them, only skipping them. There are a few programs around that are able to are able to transcode ID3v1 and ID3v2 to FLAC tags.

sezero:
Then I'd say that we don't apply this patch. @slouken, @icculus?

Wohlstand:
My patch doesn't adds support: it adds skipping as all current FLAC implementations. Without this patch, all these FLAC files gets recognised as MP3 files that obviously gets refused as invalid by MP3 decoders.

Also, recently at MixerX I slightly tweaked my code to also process the check for MP3 frames after ID3 tag to ensure that it's really an MP3 file and not something also. I could extend this patch by these changes I did.

Wohlstand:
Rebased the thing.

Wohlstand:
Just now I rebased these changes to the latest state.

icculus:
This works correctly in SDL3_mixer, and we're probably not applying this to SDL2_mixer at this point.

Wohlstand:
> This works correctly in SDL3_mixer, and we're probably not applying this to SDL2_mixer at this point.

Does this works with SDL_RWops method without trusting to filename extensions? The goal of THIS patch is make the detection by data being proper since in my projects I don't trust filename extensions and I always use data detection method, and I know what happens.

Wohlstand:
Even more, the reason why I made this patch (if not count the detection of legacy ID3-equipped FLAC files, it fixes the detection of a large amont of legit MP3 files that has various junk at begining that makes them being never identified because the legit frame doesn't begins at the begin of the file or at end of ID3 tag).

icculus:
Yes, we don't use the file extension, and we check all files for ID3, even if they shouldn't in theory be able to have one, because it catches cases like this.

Wohlstand:
> Yes, we don't use the file extension, and we check all files for ID3, even if they shouldn't in theory be able to have one, because it catches cases like this.

I took the review of SDL3_mixer and I confirm that, I am about SDL**2**_mixer for what this patch is.

Wohlstand:
> we check all files for ID3, even if they shouldn't in theory be able to have one

I may suspect, some users might use tagging software with a wrong files, and add ID3tag to file that supposed to never have it, even something unexpected.

PR #532, opened 2023-06-06
[SDL2] CMake: Improve dependency searching

FtZPetruska:
This PR is a backport of #504 for the SDL2 branch.

### DIFF
--- a/cmake/FindOpusFile.cmake
+++ b/cmake/FindOpusFile.cmake
@@ -1,19 +1,30 @@
 include(FindPackageHandleStandardArgs)
 
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_OPUSFILE QUIET opusfile)
+
 find_library(OpusFile_LIBRARY
     NAMES opusfile
+    HINTS ${PC_OPUSFILE_LIBDIR}
 )
-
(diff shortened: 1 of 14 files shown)


cmake/FindOpusFile.cmake
```diff
@@ -1,20 +1,31 @@
 include(FindPackageHandleStandardArgs)
 
+find_package(PkgConfig QUIET)
+pkg_check_modules(PC_OPUSFILE QUIET opusfile)
```
madebr, line 4: I think this must be executed conditionally if pkg-config is not installed.

FtZPetruska: It is safe to call even when the pkg-config executable is not found.

Internally, the modules guards the call with `if(PKG_CONFIG_EXECUTABLE)`, setting `PKG_CONFIG_EXECUTABLE` manually to `PKG_CONFIG_EXECUTABLE-NOTFOUND` does not break configuration, instead, it skips searching and sets all the result variables to an empty string.

madebr: Yup. [You are right](https://github.com/Kitware/CMake/blob/v3.0.0/Modules/FindPkgConfig.cmake#L263).

PR #538, opened 2023-07-27
Updated for new SDL3 audio API.

icculus:
This is a draft PR until the new audio API is merged into SDL3, but once that is, this will update SDL_mixer to work with it.

This makes two API changes to SDL mixer.

First, Mix_OpenAudio is simplified, and can even be called as `Mix_OpenAudio(0, NULL);` and probably get the exact desired result in most use cases.

Second, Mix_OpenAudioDevice is gone, as redundant.

Internally, this is the most minimal patch that gets SDL_mixer working again, which is to say it cleans up a few things that expected SDL_AudioSpec to have more fields, and it changes the library to use a single SDL_AudioStream with a callback instead of the SDL2 audio device callback.

One could certainly get more creative, turning each SDL_mixer channel into an audio stream bound to the device, each of them firing a callback that allows SDL_mixer to decode audio on the fly and perform effects, but that's something we should think about if we ever decide to do a serious SDL_mixer overhaul.

This PR is just making sure the library is still usable.

### DIFF
--- a/examples/playmus.c
+++ b/examples/playmus.c
@@ -113,10 +113,6 @@ static void IntHandler(int sig)
 
 int main(int argc, char *argv[])
 {
-    int audio_rate;
-    Uint16 audio_format;
-    int audio_channels;
-    int audio_buffers;
     int audio_volume = MIX_MAX_VOLUME;
     int looping = 0;
     int interactive = 0;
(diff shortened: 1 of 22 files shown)


icculus:
Forced-pushed, and will merge this once the GitHub Actions complete.

PR #539, opened 2023-08-01
Prevent a double call of the Mix_ChannelFinished() callback when using Mix_FadeOutChannel()

oleg-derevenetz:
Currently, in the `main` branch, in a situation when `Mix_FadeOutChannel()` is called before the end of the playback, but when there is less than `ms` milliseconds left before the end of the track, there will be two calls of `Mix_ChannelFinished()` callback instead of one - the first is when playback is finished, and the second is when the fade timer expires. This may lead to various issues (e.g. to the double-free issue when `Mix_ChannelFinished()` callback is used to release the `Mix_Chunk`).

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -406,6 +406,8 @@ mix_channels(void *udata, Uint8 *stream, int len)
 
                     /* rcg06072001 Alert app if channel is done playing. */
                     if (!mix_channel[i].playing && !mix_channel[i].looping) {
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

oleg-derevenetz:
Hi @slouken it would be nice if this commit would be eventually added to the SDL2 branch as well.

PR #547, opened 2023-08-27
[SDL2] Fixes for WAV playing

Wohlstand:
There are two bugs got been fixed:
- Added the skipping of pad byte to fix the work of several WAV files like this one:
[Cymatics_-_BELL_Dusty_C.wav.zip](https://github.com/libsdl-org/SDL_mixer/files/12448832/Cymatics_-_BELL_Dusty_C.wav.zip) that contains `bext` chunk with 603 length, but actual size of the chunk os 604 because of a pad byte.
- Fixed the memory out-of-range access at fetch_float64be() and fetch_float64le() calls.


Wohlstand:
Force-pushed to fix the build

PR #551, opened 2023-08-30
music_wav.c: Fixed ADPCM buffer overflow

### LINKED ISSUE #550: [SDL3, ADPCM] Running out of buffer at `state->output.data` buffer size at MS_ADPCM_DecodeBlockData
Recently, I tried to make a small experiment with MS ADPCM thing, and I accidentally caught an AddressSanitizer's error at the MS_ADPCM_DecodeBlockData() call. Then, I added the assertion:
```
SDL_assert(outpos < state->output.size);
```
prior to the:
```
state->output.data[outpos++] = sample1;
```
and it fails:
Снимок экрана от 2023-08-30 07-42-13

I hadn't figured yet for the source of issue yet, but if I will find something, I'll make a pull-request. Otherewise, let me know if somebody make a fix before me.

P.S. The file that causes an error:
[a_ja_tolko_s_moroza_[antonov_andrei]-adpcm_ms.wav.zip](https://github.com/libsdl-org/SDL_mixer/files/12471763/a_ja_tolko_s_moroza_.antonov_andrei.-adpcm_ms.wav.zip)

I generated it via ffmpeg somewhere in 2018 for my local tests. It works just fine at other players such as VLC.

P.P.S. Also IMA ADPCM too:
[a_ja_tolko_s_moroza_[antonov_andrei]-adpcm_ima_wav.wav.zip](https://github.com/libsdl-org/SDL_mixer/files/12472123/a_ja_tolko_s_moroza_.antonov_andrei.-adpcm_ima_wav.wav.zip)

Wohlstand:
I accidentally found that the output buffer was been created for mono output only, and therefore, on attempt to play stereo files, the buffer overflow gets occured. So, the size of the buffer must be multipled by a number of channels.

The test file that causes a problem, was attatched to the referred issue.

Fixes #550

### DIFF
--- a/src/codecs/music_wav.c
+++ b/src/codecs/music_wav.c
@@ -561,7 +561,7 @@ static int MS_ADPCM_Init(ADPCM_DecoderState *state, const Uint8 *chunk_data, Uin
 
     state->output.read = 0;
     state->output.pos = 0;
-    state->output.size = state->samplesperblock;
+    state->output.size = state->samplesperblock * state->channels;
     state->output.data = (Sint16 *)SDL_malloc(state->output.size * sizeof(Sint16));
     if (!state->output.data) {
         return Mix_OutOfMemory();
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

Issue #542, opened 2023-08-09
lots of "defined in the headers but not the wiki!" warnings


sezero:
E.g.: `WARNING: Mix_AllocateChannels defined in the headers but not the wiki!`
Seen in the CI logs too.

slouken:
I think we need someone to go through and update the wiki with all of our changes.

Any volunteers?

madebr:
These are caused by this line:
https://github.com/libsdl-org/SDL_mixer/blob/14f1eb25603cc75624e9d2c92a5b0eebcf93a015/.wikiheaders-options#L15
It is set to 0 in SDL, but to 1 in the satellite libraries.

I asked here: https://github.com/libsdl-org/SDL/issues/7514#issuecomment-1608560156 (but I don't think it got answered)

madebr:
> I think we need someone to go through and update the wiki with all of our changes.
> 
> Any volunteers?

The manpage generator script always generates the wiki/man pages from scratch.
So when first generating the docs, all pages are guaranteed to be missing.

slouken:
Oh, okay. So is it fine just to remove the warning?

madebr:
I think the warning makes sense when updating the sdl wiki, but not when generating the headers.
So perhaps it should be changed into a cli argument-only option.

sezero:
https://github.com/libsdl-org/SDL_mixer/pull/552 seems to have fixed this.

sezero:
BTW, the SDL2 branch has the same `warn_about_missing = 1` but I don't remember getting the warnings.

PR #552, opened 2023-09-04
wikiheaders: don't warn about missing files by default

madebr:
https://github.com/libsdl-org/SDL_mixer/issues/542

`wikiheaders.pl` has [`--warn-about-missing`](https://github.com/libsdl-org/SDL_mixer/blob/489716b38b7293f2539f37714dc0890cbace37c5/build-scripts/wikiheaders.pl#L36) to force the value to 0.

@icculus Is this ok? Or is a `--no-warn-about-missing` option preferred?

### DIFF
--- a/.wikiheaders-options
+++ b/.wikiheaders-options
@@ -12,5 +12,5 @@ selectheaderregex = \ASDL_mixer\.h\Z
 projecturl = https://libsdl.org/projects/SDL_mixer
 wikiurl = https://wiki.libsdl.org/SDL_mixer
 bugreporturl = https://github.com/libsdl-org/sdlwiki/issues/new
-warn_about_missing = 1
+warn_about_missing = 0
 wikipreamble = (This function is part of SDL_mixer, a separate library from SDL.)


slouken:
Suggest removing "about", e.g. `--warn-missing` or `--warn-missing-files`

madebr:
> Suggest removing "about", e.g. `--warn-missing` or `--warn-missing-files`

That's up to @icculus. In a sense, the option is private, and meant for project maintainers only.

icculus:
Let's either fix the missing things or change the default in the script.

madebr:
> Let's either fix the missing things or change the default in the script itself.

[The warning is emitted when a function is documented in the wiki but not in the headers (and the reverse case)](https://github.com/libsdl-org/SDL_mixer/blob/dc93c43b4d58f373059ba4811abf99e5c8fde33c/build-scripts/wikiheaders.pl#L792-L806)
This warning makes sense when generating the wiki at https://wiki.libsdl.org, but does not when generating the man pages.
(the perl script cannot generate man pages without the wiki pages present)

icculus:
Really it doesn't make sense when generating wiki.libsdl.org either, because the cronjob just sends it to me in an email where I ignore it.  :) 

What if, instead of printing warnings, it generates a wiki page called "Undocumented.md" that updates on each run? That way we would have a public list of things that lack documentation.

icculus:
For now, though, let's just change the default in the script.

madebr:
> For now, though, let's just change the default in the script.

The default is fine:

https://github.com/libsdl-org/SDL_mixer/blob/3cb07a1f4b6fa02b83298c025f2994cc770cbcf0/build-scripts/wikiheaders.pl#L28

It's the project-specific configuration that has the override:

https://github.com/libsdl-org/SDL_mixer/blob/3cb07a1f4b6fa02b83298c025f2994cc770cbcf0/.wikiheaders-options#L15

[SDL has it disabled](https://github.com/libsdl-org/SDL/blob/f2ca9a615b3572368641daeef7bdb3ff0688a0dd/.wikiheaders-options#L16)

icculus:
Oh. Ok then.  :)

madebr:
> What if, instead of printing warnings, it generates a wiki page called "Undocumented.md" that updates on each run? That way we would have a public list of things that lack documentation.


Great idea.
About wikiheaders.pl, in case you missed it, I pushed this patch yesterday: https://github.com/libsdl-org/SDL/commit/e85206ffd8becd33321182614e2ba5a5d304a5a3

PR #554, opened 2023-09-05
cmake: allow building SDL_mixer with not all dependencies present + make libflac/drflac mutually exclusive

madebr:
- When CMake is configured with `-DSDL3MIXER_STRICT=OFF` , you can build SDL3_mixer with some dependencies not present (that feature will be disabled)
- Make drflac and libflac mutually exclusive. By default, drflac is used.


madebr:
I accidentally pushed this to libsdl-org/SDL_mixer.

PR #555, opened 2023-09-05
cmake: allow building SDL_mixer with not all dependencies present + disable libflac by default

madebr:
- When CMake is configured with `-DSDL3MIXER_STRICT=OFF` , you can build SDL3_mixer with some dependencies not present (that feature will be disabled). The default is `-DSDL3MIXER_STRICT=OFF`.
- ~Make drflac and libflac mutually exclusive. By default, drflac is used.~ Disable libflac by default, keep drflac enabled.

fixes #553

PR #557, opened 2023-09-19
build: Fix detection of libvorbisfile

smcv:
The variable name is case-sensitive and the module is named `Vorbis`, not `vorbis`.


smcv:
I noticed this while updating the snapshots in Debian experimental: `playmus` couldn't play an Ogg Vorbis music track, because we were disabling all backends except `vorbisfile`, but then wrongly thinking `libvorbisfile` hadn't been found. This PR fixes that.

We're building with `-DSDL3MIXER_STRICT=ON`, so if we genuinely hadn't found libvorbisfile the build would have failed; but `-DSDL3MIXER_STRICT=ON` didn't catch this particular failure mode. Maybe this means more `message()` calls should have the `${fatal_error}` parameter?

sezero (approved):
LGTM. @madebr?

sezero:
P.S.: SDL2 affected?

smcv:
> P.S.: SDL2 affected?

I don't know, I'm using Autotools in the Debian packaging of all of the SDL2 family (and will continue to do so unless told that upstream maintainers are no longer recommending it).

The CMakeLists.txt in the SDL2 branch looks different enough that I think it's probably unaffected.

madebr:
The fix is correct!

I will replace some `STATUS`es with `${fatal_error}` as a fail safe.
I was assuming `find_package(REQUIRED)` would be enough for the strict mode,
but I did not anticipate cmake bugs :).

SDL2 does not have this issue.
The strict mode is introduced in https://github.com/libsdl-org/SDL_mixer/pull/555, which is SDL3 only.

smcv:
> I was assuming `find_package(REQUIRED)` would be enough for the strict mode, but I did not anticipate cmake bugs

I don't think this is a CMake bug, only a bug in our `CMakeLists.txt`. In my build, `find_package(Vorbis ${required})` correctly detected the libvorbis family and set `Vorbis_FOUND` (so the `REQUIRED` didn't have any practical effect, because the required thing was found). But then the `CMakeLists.txt` looked at `vorbis_FOUND` instead, and that variable wasn't set, leading to this:

```
-- Found Vorbis: /usr/lib/x86_64-linux-gnu/libvorbisfile.so¯¯                     
-- vorbisfile NOT found                                                           
```

madebr:
That's what I meant: a bug in my cmake code.
The `${fatal_error}` of my latest patch should catch this error case.
Thanks again!

PR #559, opened 2023-09-20
Add Mix_ChannelFinished to SDL_mixer.sym

ohthesetrees:
Adds a function to SDL_mixer.sym that was previously missed and was causing me some linker errors.


slouken:
Merged, thanks!

PR #560, opened 2023-09-20
Add Mix_PauseGroup and Mix_ResumeGroup functions

ohthesetrees:
Adds a couple helper functions similar to `Mix_HaltGroup` that make it easy to pause and resume groups of channels together. In my game I group in-game sounds so I can easily pause and resume them when the game pauses.


slouken:
This seems like a great addition, thanks!

PR #562, opened 2023-10-08
Fix some \param comments to match real parameter names/list

JLKenyon:


madebr:
Thanks you! I backported it to SDL2 in fef85f45

PR #566, opened 2023-10-11
Fix flac-in-ogg handling

ankith26:
The first commit of this PR is for fixing the detection of FLAC files when it is encapsulated in an OGG transport later (currenty these kind of files go down the Ogg-Vorbis path, which cannot handle it).

The second commit fixes the handling of such files in the libflac backend (the drflac backend seems to already take care of this).

I wrote and tested these commits against the SDL2 branch. With few small porting changes I'm making this PR against `main`. If the maintainers wish I'd be happy to also open the SDL2 branch backport PR for this.

### DIFF
--- a/src/codecs/music_flac.c
+++ b/src/codecs/music_flac.c
@@ -49,6 +49,17 @@ typedef struct {
                         FLAC__StreamDecoderMetadataCallback metadata_callback,
                         FLAC__StreamDecoderErrorCallback error_callback,
                         void *client_data);
+    FLAC__StreamDecoderInitStatus (*FLAC__stream_decoder_init_ogg_stream)(
+                        FLAC__StreamDecoder *decoder,
+                        FLAC__StreamDecoderReadCallback read_callback,
+                        FLAC__StreamDecoderSeekCallback seek_callback,
+                        FLAC__StreamDecoderTellCallback tell_callback,
+                        FLAC__StreamDecoderLengthCallback length_callback,
+                        FLAC__StreamDecoderEofCallback eof_callback,
+                        FLAC__StreamDecoderWriteCallback write_callback,
+                        FLAC__StreamDecoderMetadataCallback metadata_callback,
+                        FLAC__StreamDecoderErrorCallback error_callback,
+                        void *client_data);
     FLAC__bool (*FLAC__stream_decoder_finish)(FLAC__StreamDecoder *decoder);
     FLAC__bool (*FLAC__stream_decoder_flush)(FLAC__StreamDecoder *decoder);
     FLAC__bool (*FLAC__stream_decoder_process_single)(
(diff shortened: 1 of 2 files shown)


sezero:
I'm not against this, but not sure how common are the ogg-flac files are.

In any case, shouldn't we handle the extension in `Mix_LoadMUS` too? IIRC,
ogg-flac files usually have `oga` extension.

ankith26:
I'm not sure we can assume every `.oga` file is a flac-in-ogg. There can be even more obscure formats that use ogg as the container format. I think it's instead better to go down the detector path instead of making incorrect guesses

sezero:
OK then. @slouken and/or @icculus: Please review + decide for this one.

sezero (approved):

sezero:
> OK then. @slouken and/or @icculus: Please review + decide for this one.

@slouken: PING: Do we want this one?

src/codecs/music_flac.c
```diff
@@ -489,6 +511,14 @@ static void *FLAC_CreateFromRW(SDL_RWops *src, SDL_bool freesrc)
...
     int was_error = 1;
     FLAC__int64 full_length;
+    int is_ogg_flac;
+    Uint8 magic[4];
+    if (SDL_RWread(src, magic, 4) != 4) {
+        SDL_SetError("Couldn't read first 4 bytes of audio data");
+        return NULL;
+    }
+    SDL_RWseek(src, -4, SDL_RW_SEEK_CUR);
+    is_ogg_flac = !SDL_memcmp(magic, "OggS", 4);
```
slouken, line 521: This shouldn't be treated as a boolean, instead please do (SDL_memcmp(magic, "OggS", 4) == 0)

src/codecs/music_flac.c
```diff
@@ -500,17 +530,30 @@ static void *FLAC_CreateFromRW(SDL_RWops *src, SDL_bool freesrc)
...
-                    music) == FLAC__STREAM_DECODER_INIT_STATUS_OK) {
+        if (is_ogg_flac) {
+            ret = flac.FLAC__stream_decoder_init_ogg_stream(
+                music->flac_decoder,
+                flac_read_music_cb, flac_seek_music_cb,
+                flac_tell_music_cb, flac_length_music_cb,
+                flac_eof_music_cb, flac_write_music_cb,
+                flac_metadata_music_cb, flac_error_music_cb,
+                music);
+        }
```
slouken, line 546: SDL style is else on the same line: "} else {"

slouken:
Conceptually this is great. I added a couple of style notes and we can merge this when they're complete.

ankith26:
I have fixed formatting issues. Also let me know if I can open a backport to SDL2 for this PR

sezero:
This is in, thanks!

>  if I can open a backport to SDL2 for this PR

Please do. (It's only different in two lines..)

sezero:
> > if I can open a backport to SDL2 for this PR
> 
> Please do. (It's only different in two lines..)

Well, I went ahead and cherry-picked it to SDL2 branch as https://github.com/libsdl-org/SDL_mixer/commit/cef595086bf20ca12b31d591fbaaf212721d0b81
Please confirm.

ankith26:
Yup, that looks good to me

PR #575, opened 2023-12-01
Fix libxmp-lite pkgconf

sharkwouter:
Without this change I get the following error when using pkg-config after building with libxmp-lite:
```
Package libxmplite was not found in the pkg-config search path.
Perhaps you should add the directory containing `libxmplite.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libxmplite', required by 'SDL2_mixer', not found
```
This is mostly relevant for Playstation Portable and Vita builds, but there are probably other platforms which want support for the formats libxmp-lite offers. Especially since is has an easier to work with license than libxmp.


CMakeLists.txt
```diff
@@ -687,10 +687,10 @@ if(SDL2MIXER_MOD_XMP)
         if(SDL2MIXER_MOD_XMP_LITE)
             message(STATUS "Using system libxmp-lite")
             find_package(libxmp-lite REQUIRED)
-            set(tgt_xmp libxmp-lite::libxmp-lite)
+            set(tgt_xmp libxmp-lite)
```
madebr, line 690: The cmake scripts of [latest libxmp-lite](https://github.com/libxmp/libxmp/releases/tag/libxmp-4.6.0) creates `libxmp-lite::xmp_lite_shared` and `libxmp-lite::xmp_lite_static` targets.
[libxmp-lite 4.5.0](https://github.com/libxmp/libxmp/releases/tag/libxmp-4.5.0) does not have any cmake support.

[The vita build file](https://github.com/vitasdk/packages/blob/master/libxmp-lite/VITABUILD#L5) looks normal.
I don't see how its `*config.cmake` creates a `libxmp-lite` target.

What about the following?
```suggestion
            set(tgt_xmp libxmp-lite::llbxmp-lite::xmp_lite_shared)
            if(NOT TARGET "${tgt_xmp}")
                set(tgt_xmp libxmp-lite::llbxmp-lite::xmp_lite_static)
            endif()
```

Perhaps `libxmp-lite`'s cmake script should create a default `libxmp-lite::libxmp-lite` target, like SDL does (`SDL3::SDL3-shared`/`SDL3::SDL3-static` => `SDL3::SDL3`) [here](https://github.com/libsdl-org/SDL/blob/058213366bccf7f62d4c4c6061477d7a69ea336d/cmake/SDL3Config.cmake.in#L79-L96)

sharkwouter: Thanks, I'll test this

sharkwouter: I've done some testing and found that the solution which I proposed here in the PR works and the change suggested results in the following error while running cmake:
```
CMake Error at CMakeLists.txt:216 (add_library):
  Target "SDL2_mixer" links to target "libxmp-lite::xmp_lite_static" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```
My guess is that either a libxmp-lite build using cmake is different than one using autotools or SDL3 is also unable to build with libxmp-lite.

sharkwouter: I did fix the typo in the suggestion when testing it. it says lbb instead of lib in 2 places.

madebr: I just tried it in a vita docker container, and current sdl2 master works.
```
$ docker run --rm -ti -v ~/projects/SDL_mixer:/sdl2_mixer:z  vitasdk/vitasdk
```
```
$ apk update
$ apk add cmake ninja pkgconf
$ vdpm libxmp-lite
$ cmake -S /sdl2_mixer/ -B /sdl2_mixer-build -GNinja -DCMAKE_INSTALL_PREFIX=/sdl2_mixer-prefix -DCMAKE_TOOLCHAIN_FILE=${VITASDK}/share/vita.toolchain.cmake -DSDL2MIXER_MOD_XMP=ON -DSDL2MIXER_MOD_XMP_LITE=ON -DSDL2MIXER_MOD=ON -DBUILD_SHARED_LIBS=OFF -DSDL2MIXER_DEPS_SHARED=OFF -DSDL2MIXER_OPUS_SHARED=OFF -DSDL2MIXER_MOD_XMP_SHARED=OFF -DSDL2MIXER_MIDI_FLUIDSYNTH=OFF -DSDL2MIXER_WAVPACK=OFF
$ cmake --build --build /sdl2_mixer-build
```

`playmus` was linked with the following command, correctly containing `libxmp-lite.a`:
```
/usr/local/vitasdk/bin/arm-vita-eabi-gcc -Wl,-q -Wl,-z,nocopyreloc CMakeFiles/playmus.dir/playmus.c.obj -o playmus  libSDL2_mixer.a  /usr/local/vitasdk/arm-vita-eabi/lib/libSDL2main.a  /usr/local/vitasdk/arm-vita-eabi/lib/libSDL2.a  -lSceGxm_stub  -lSceDisplay_stub  -lSceCtrl_stub  -lSceAppMgr_stub  -lSceAppUtil_stub  -lSceAudio_stub  -lSceAudioIn_stub  -lSceSysmodule_stub  -lSceIofilemgr_stub  -lSceCommonDialog_stub  -lSceTouch_stub  -lSceHid_stub  -lSceMotion_stub  -lScePower_stub  -lSceProcessmgr_stub  /usr/local/vitasdk/arm-vita-eabi/lib/libopusfile.a  -logg  -lopus  -lm  /usr/local/vitasdk/arm-vita-eabi/lib/libxmp-lite.a &&
```

sharkwouter: I'll do another test without the target change.

sharkwouter: Seems I got confused and made a change which was indeed not needed. I'm sorry for the time it took for me to figure that out. I've undone that part and only left the part that is required for things to work again.

madebr (approved):
https://github.com/libxmp/libxmp/blob/master/lite/libxmp-lite.pc.in

madebr:
main needs the same fix.
Feel free to open a PR, or I'll do it once I get back to my PC.
https://github.com/libsdl-org/SDL_mixer/blob/76720e6ea298f937548e54f75083133bf4c86f1b/CMakeLists.txt#L821

madebr:
Applied to SDL3_mixer in 1d30b089adf8eedeb0559421619171a610948216

PR #584, opened 2024-01-14
minimp3: remove TAG+ support

sezero:
It's a still-born extension that no one uses, no player supports, and no tag library parses.

And it's not impossible to hit it as a false positive..

(The removal has already been suggested to mainstream at https://github.com/lieff/minimp3/issues/66)

---

If https://github.com/mackron/dr_libs/issues/263 doesn't get  resolved and we won't go back to dr_mp3, please merge this and also cherry-pick to SDL3 branch as well.


slouken:
I went ahead and merged and cherry-picked.

PR #586, opened 2024-01-15
wavpack: WavpackOpenFileInput is not used, so don't load it

### LINKED ISSUE #585: wavpack misses ftruncate
Building SDL_mixer's playmus for ps2 and psp (on ci), linking fails with:
```
  /tmp/setupsdl/da8de0a910961c86f12e0e6d22427875f5c38084675d3be163533c2ff1367bcc/source/external/wavpack/src/open_filename.c:208:(.text+0x40): undefined reference to `ftruncate'
  collect2: error: ld returned 1 exit status
```

This probably means wavpack is missing a link to the math library.
Or the toolchain does not provide this symbol.

madebr:
`ftruncate` is used through`WavpackOpenFileInput`.
Since SDL_mixer uses `WavpackOpenFileInputEx` instead of `WavpackOpenFileInput`, don't look for `WavpackOpenFileInput`.

Depends on https://github.com/libsdl-org/WavPack/pull/2

Fixes #585

### DIFF
--- a/src/codecs/music_wavpack.c
+++ b/src/codecs/music_wavpack.c
@@ -66,7 +66,6 @@ typedef struct {
     uint32_t libversion;
     uint32_t (*WavpackGetLibraryVersion)(void);
     char *(*WavpackGetErrorMessage)(WavpackContext*);
-    WavpackContext *(*WavpackOpenFileInput)(const char *infilename, char *error, int flags, int norm_offset);
     WavpackContext *(*WavpackOpenFileInputEx)(WavpackStreamReader *reader, void *wv_id, void *wvc_id, char *error, int flags, int norm_offset);
     WavpackContext *(*WavpackCloseFile)(WavpackContext*);
     int (*WavpackGetMode)(WavpackContext*);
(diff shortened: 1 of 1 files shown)


src/codecs/music_wavpack.c
```diff
@@ -66,7 +66,8 @@ typedef struct {
     uint32_t libversion;
     uint32_t (*WavpackGetLibraryVersion)(void);
     char *(*WavpackGetErrorMessage)(WavpackContext*);
-    WavpackContext *(*WavpackOpenFileInput)(const char *infilename, char *error, int flags, int norm_offset);
+    /* WavpackOpenFileInput is not supported on all platforms (psp and ps2 are missing ftruncate) */
+    /* WavpackContext *(*WavpackOpenFileInput)(const char *infilename, char *error, int flags, int norm_offset); */
```
sezero, line 70: Just remove it instead of commenting it out: I can't remember why I included that function at all (probably is a remnant of some early tests of mine..)

madebr: And remote the comment as well?

sezero: Yep

sezero:
> Depends on [libsdl-org/WavPack#2](https://github.com/libsdl-org/WavPack/pull/2)

It will be a problem if you rebase that PR: you should manually merge it
like:
```
git checkout 5.6.0-sdl && \
git pull https://github.com/madebr/WavPack.git no-WavpackOpenFileInput-for-ps2-and-psp && \
git push
```
... so that that the relevant commit is in our branch.

madebr:
Indeed, it will be a problem when it is merged through the github ui.
If you think the patch is correct, I can just push to libsdl-org/wavpack directly.

madebr:
I applied it to SDL2_mixer in 40b5164882aa0aed8db47e68a4bf5ea3f24396b8 (I forgot to add `-x`)

PR #594, opened 2024-02-23
examples/playmus: Prevent stack smashing on long input strings

tatokis:
In addition, fflush(stdin) is undefined behaviour, so remove it.


sezero:
Sensible I guess. @slouken?

slouken:
Merged, thanks!

PR #615, opened 2024-05-26
native_midi_common.c (ReadMIDIFile): fix possible NULL ptr dereference.

sezero:
if allocation of mididata->track ever fails we'd still be dereferencing it under 'bail'

Unless I'm missing something...


slouken (approved):
This looks good, thanks!

PR #620, opened 2024-07-04
Mix_AllocateChannels realloc fix

zanadoman:
Hi! I recently noticed that Mix_AllocateChannels didn't have a useful return value, it just returned the same number of channels that I passed in as parameter. Therefore I found out that there is not a single error checking on the success of the allocation, and in the case of an unsuccessful allocation an invalid address would have been used.
Solution: I added the missing error checking, the reallocation now uses a temporary pointer to check the success of the allocation, if the allocation was successful it will use the new pointer otherwise it sets an error, and the original pointer remains intact. Also now you can check the return value for useful information on the success of the allocation.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -575,6 +575,8 @@ void Mix_PauseAudio(int pause_on)
  */
 int Mix_AllocateChannels(int numchans)
 {
+    struct _Mix_Channel *mix_channel_tmp;
+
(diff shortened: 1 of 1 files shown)


zanadoman:
Also added a case for 0 numchans because realloc(x, 0) is not equivalent to free(x) and valgrind detected the error.

slouken:
Merged, thanks!

PR #632, opened 2024-09-05
cmake: allow overriding GME_ZLIB and GME_UNRAR options by user

mechakotik:
I am using vendored SDL_mixer in my project and I need zlib support in libgme to load compressed VGMs. Current SDL_mixer CMake config disables it and doesn't allow user to enable it.


CMakeLists.txt
```diff
@@ -701,8 +701,12 @@ if(SDLMIXER_GME)
...
         set(ENABLE_UBSAN OFF)
         set(BUILD_FRAMEWORK OFF)
-        set(GME_ZLIB OFF)
-        set(GME_UNRAR OFF)
+        if(NOT DEFINED GME_ZLIB)
+            set(GME_ZLIB OFF)
+        endif()
+        if(NOT DEFINED GME_UNRAR)
+            set(GME_UNRAR OFF)
+        endif()
```
madebr, line 709: Would this work for you too?
```suggestion
        set(GME_ZLIB "OFF" CACHE BOOL "Build GME with zlib support")
        set(GME_UNRAR "OFF" CACHE BOOL "Build GME with unrar support")
```

mechakotik: It doesn't work for me. I am setting GME_ZLIB from the project CMakeLists.txt before including SDL_mixer.

madebr: You need to do the following in your cmake script:
```cmake
set(GME_ZLIB ON CACHE BOOL "GME with zlib" FORCE)
set(GME_UNRAR ON CACHE BOOL "GME with unrar" FORCE)
```

mechakotik: Now it works, but as user just writing:
```cmake
set(GME_ZLIB ON)
```
seems like more intuitive way for me.

madebr: This one has a better behavior, I think.
```suggestion
        option(GME_ZLIB "GME with zlib" OFF)
        option(GME_UNRAR "GME with unrar" OFF)
```

mechakotik: Yes, this seems like a better way. Also libgme CMakeLists no longer has GME_UNRAR option.

madebr (approved):

mechakotik:
Could you please merge it into SDL2 too?

madebr:
Sure, see f66ffea0 (SDL2) and 33668fa3c2e78afe71493aea09ca1d24d4c3dcfe (2.8.x)

PR #636, opened 2024-09-20
Android.mk: add GME support

mechakotik:


Android.mk
```diff
@@ -136,6 +145,12 @@ ifeq ($(SUPPORT_WAVPACK),true)
     LOCAL_STATIC_LIBRARIES += wavpack
 endif
 
+ifeq ($(SUPPORT_GME),true)
+    LOCAL_CFLAGS += -DMUSIC_GME
+    LOCAL_C_INCLUDES += $(LOCAL_PATH)/$(GME_LIBRARY_PATH)
```
sezero, line 150: Do you not need to append `/gme` here ?

mechakotik: SDL_mixer source includes it like `#include <gme/gme.h>`, so it's correct include path.

sezero: Ah yes. Sorry for the noise.

sezero:
Patch is in, and cherry-picked to SDL2 branch too. Thanks.

PR #637, opened 2024-09-24
Linux/ALSA Native MIDI backend

fragglet:
This (finally!) adds a native MIDI backend for Linux using the ALSA sequencer interface. I've tested this with Chocolate Doom and the Timidity++ ALSA daemon and it works pretty reliably; I haven't gotten around to testing with actual hardware yet. There are still a couple of features missing but it should be complete enough now for people to start using it.


sezero:
This was somehow missed and #595 was merged...

@fragglet: Can we pick bits from here into presently existing one? If yes, rebase?

CC: @slouken, @icculus

slouken:
Yes, this should go in SDL3, not SDL2.

fragglet:
I'm fine with us going with the #595 route instead. Just really happy to see SDL_mixer get a Linux midi backend at last!

If there is anything from my implementation that look like it's missing from the one that got merged, please let me know and I can adapt it.

sezero:
> I'm fine with us going with the #595 route instead. Just really happy to see SDL_mixer get a Linux midi backend at last!
> 
> If there is anything from my implementation that look like it's missing from the one that got merged, please let me know and I can adapt it.

CC @tatokis

tatokis:
I don't immediately see anything missing. One thing I am unsure about is Sysex handling, as I don't think my implementation passes them through. I'll look into it shortly.

sezero:
> I don't immediately see anything missing. One thing I am unsure about is Sysex handling, as I don't think my implementation passes them through. I'll look into it shortly.

Great, thanks

fragglet:
> I don't immediately see anything missing. One thing I am unsure about is Sysex handling, as I don't think my implementation passes them through. I'll look into it shortly.

I have a personal interest in making sure sysex messages are handled properly (one of my original motivations was to be able to hook up synths like the Roland sc55) so if you don't have the time to look into it please say and I can take it off your hands

tatokis:
> I have a personal interest in making sure sysex messages are handled properly (one of my original motivations was to be able to hook up synths like the Roland sc55) so if you don't have the time to look into it please say and I can take it off your hands

Please have a look at https://github.com/libsdl-org/SDL_mixer/pull/679 and let me know if it works.

PR #642, opened 2024-10-13
Don't statically link libxmp

ankith26:
The `libxmp` dependency seems to be inconsistently handled when building with `SDL2MIXER_DEPS_SHARED=0`. Every other library uses "standard dynamic linking" instead of the "SDL style dynamic linking" (apologies if this is wrong terminology, but I don't know what else to call it). But `libxmp` does an actual static linking, which increases the size of `SDL_mixer` shared library. I would expect it to do the "standard dynamic linking" when I pass `SDL2MIXER_DEPS_SHARED=0`.

This is a cmake only issue, the autotools build handles this as expected. This fix if accepted, needs to be frontported to SDL3 branch as well I believe.

PR #644, opened 2024-10-19
[SDL3 frontport] Don't statically link libxmp

ankith26:
SDL3 frontport of #642

Should I make a similar PR against the 2.8.x branch as well?


sezero:
> Should I make a similar PR against the 2.8.x branch as well?

I think that old release branch should be left alone.

slouken:
Merged, thanks!

PR #653, opened 2024-12-22
Convert volume API from integers [0-128] to floats [0-1]

### LINKED ISSUE #652: SDL3: update volume to a float
In SDL2, SDL_mixer uses https://wiki.libsdl.org/SDL2/SDL_MixAudioFormat, which takes a volume from 0-128.

In SDL3, SDL_mixer uses https://wiki.libsdl.org/SDL3/SDL_MixAudio, which now takes a float.

According to SDL's README-MIGRATION:

> The following symbols have been removed:
> SDL_MIX_MAXVOLUME - mixer volume is now a float between 0.0 and 1.0

Therefore, I think it would make sense for SDL_mixer to follow SDL's lead here, before the API is frozen for release.

Starbuck5:
In a lot of places this actually makes the code simpler. Volumes can be multiplied without then dividing by the integer range, the new SDL3 mixing API can be used without another divide into float-land, it makes one or two music codecs more straightforward.

Closes #652 

A couple things I wanted to highlight in review:
- The handling of gain in music_gme.c might not be 100% the same, I'm finding it difficult to think through the x * y vs floor(x * y + 0.5)/128.
- The Timidity_SetVolume function previously took in values between 0 and 128, but compares to a MAX AMPLIFICATION of 800, and checks if the value is at least 0. That is strange.
- Is my handling of master_volume in mixer.c okay? It previously was an SDL_AtomicInt, but there is no SDL_AtomicFloat so I gave it a spinlock instead.

### DIFF
--- a/src/codecs/timidity/playmidi.c
+++ b/src/codecs/timidity/playmidi.c
@@ -785,7 +785,7 @@ int Timidity_PlaySome(MidiSong *song, void *stream, Sint32 len)
 void Timidity_SetVolume(MidiSong *song, float fvolume)
 {
   int i;
-  int volume = (int)(fvolume / 128);
+  int volume = (int)(fvolume * 128);
 
   if (volume > MAX_AMPLIFICATION)
     song->amplification = MAX_AMPLIFICATION;
--- a/src/codecs/timidity/playmidi.c
+++ b/src/codecs/timidity/playmidi.c
@@ -782,9 +782,11 @@ int Timidity_PlaySome(MidiSong *song, void *stream, Sint32 len)
   return samples * bytes_per_sample;
 }
 
-void Timidity_SetVolume(MidiSong *song, int volume)
+void Timidity_SetVolume(MidiSong *song, float fvolume)
 {
   int i;
+  int volume = (int)(fvolume / 128); // How should this be handled?
+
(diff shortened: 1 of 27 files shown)


src/codecs/timidity/playmidi.c
```diff
@@ -782,9 +782,11 @@ int Timidity_PlaySome(MidiSong *song, void *stream, Sint32 len)
   return samples * bytes_per_sample;
 }
 
-void Timidity_SetVolume(MidiSong *song, int volume)
+void Timidity_SetVolume(MidiSong *song, float fvolume)
 {
   int i;
+  int volume = (int)(fvolume / 128);
```
mkunkacnc, line 788: int volume went from 0 - 128 before so should use fvolume * 128, not fvolume / 128

Starbuck5: Great point! Thanks for pointing out.

Starbuck5: Great point! Thanks for pointing out.

Starbuck5:
Coming back to this, I see it will be overcome-by-events by icculus's SDL_remixer, where volume is controlled by a \*float\* gain. So I will close this out.

PR #654, opened 2024-12-27
Removed Mix_LockAudio and made some minor edits

edwardgushchin:
Mix_LockAudio is mentioned several times in the method documentation, but the method itself is nowhere to be found. I suspect that it used to be there, but then it was removed.


slouken:
Merged, thanks!

PR #666, opened 2025-02-13
Remove (int > uint) comparison warning in stb_vorbis.h

Slavko-P:


slouken:
Please submit as an upstream PR: https://github.com/nothings/stb

sezero:
> Please submit as an upstream PR: https://github.com/nothings/stb

That code is not in mainstream

slouken:
Ah, okay. Merged, thanks!

Slavko-P:
Now that I look at the code again:
`if (f->current_playback_loc > (int)lgs && lgs > 0 && lgs != SAMPLE_unknown)`
I think it should be:
`if (lgs > 0 && lgs != SAMPLE_unknown && f->current_playback_loc > (int)lgs)`
So the if statement doesn't even bother doing the `f->current_playback_loc > (int)lgs` comparison if lgs isn't valid.

slouken:
> Now that I look at the code again: `if (f->current_playback_loc > (int)lgs && lgs > 0 && lgs != SAMPLE_unknown)` I think it should be: `if (lgs > 0 && lgs != SAMPLE_unknown && f->current_playback_loc > (int)lgs)` So the if statement doesn't even bother doing the `f->current_playback_loc > (int)lgs` comparison if lgs isn't valid.

It's all ANDed, so it doesn't matter, but that does make sense to me. Want to make another PR?

Issue #670, opened 2025-02-19
Incredibly slow OGG file loading times when combined with PhysFS


bjorn:
I've noticed a call to `Mix_LoadMUS_RW` could take a few hundred milliseconds when called with an OGG file that is being loaded from a ZIP file through PhysFS.

While debugging this, I found it was specific to the `stb_vorbis` backend. The call takes less than a millisecond when using libvorbisfile!

The major difference appears to be because:

* libvorbisfile will read the OGG file in chunks of 2048 bytes.
* stb_vorbis appears to read the OGG file _one byte at a time_.

The latter causes millions of single-byte roundtrips through `SDL_RWops` to `PHYSFS_readBytes`.

I have not tried yet whether the situation is different in SDL_mixer 3, but the responsible code looks largely the same:

https://github.com/libsdl-org/SDL_mixer/blob/bc9c365b86734d4da667b4279faba8721c43ba06/src/codecs/stb_vorbis/stb_vorbis.h#L1398-L1403

The `get8` function has been patched up to call `SDL_ReadIO` for a single byte each time, and it appears stb_vorbis is really happy to call `get8` all over the place (even, for example, 4 times in a row for `get32`). The `getn` function, which could be more efficient, is used very sparsely.

Since STB appears to be the default library used for OGG files, this also affects my game when running on Fedora for example, so it's not really an option for me to just use SDL_mixer with libvorbisfile instead. I have found a workaround in loading the entire file in memory and using `SDL_RWFromConstMem`, which makes it much faster but it's not ideal (and introduces unnecessary overhead when SDL_mixer is using libvorbisfile).

Would it be feasible to patch up stb_vorbis to either make more use of `getn` or possibly to use a buffer like libvorbisfile is doing, to reduce the `SDL_ReadIO` calls?

sezero:
Do the following get32() changes help?

For SDL3 (main branch):
```diff
diff --git a/src/codecs/stb_vorbis/stb_vorbis.h b/src/codecs/stb_vorbis/stb_vorbis.h
index 92e5441..9437d38 100644
--- a/src/codecs/stb_vorbis/stb_vorbis.h
+++ b/src/codecs/stb_vorbis/stb_vorbis.h
@@ -1420,12 +1420,18 @@ static uint8 get8(vorb *z)
 
 static uint32 get32(vorb *f)
 {
+   #ifdef STB_VORBIS_SDL
+   Uint32 x;
[...]

PR #671, opened 2025-02-21
stb_vorbis: Use a buffer to cache SDL_RWops reads

bjorn:
Requesting the data one byte at a time can be prohibitively expensive, especially when the SDL_RWops isn't backed by constant memory but routed through PhysicsFS for example.

This change introduces a 2kb buffer, which feels small enough to not add too much overhead and is the same size as used by libvorbisfile. In my measurements, using a larger buffer no longer made a significant difference when starting to play an OGG file.

Testing with an uncompressed OGG file read from zip file, timing the call to `Mix_LoadMUS_RW`:

* No buffer: ~ 47 ms
* 10 bytes: ~ 5.2 ms
* 100 bytes: ~ 1.1 ms
* 1 kb: ~ 0.62 ms
* 2 kb: ~ 0.6 ms
* 8 kb: ~ 0.58 ms

When using the libvorbisfile backend it takes about 0.2 ms.

See #670.

### DIFF
--- a/src/codecs/stb_vorbis/stb_vorbis.h
+++ b/src/codecs/stb_vorbis/stb_vorbis.h
@@ -308,6 +308,7 @@ extern stb_vorbis * stb_vorbis_open_file_section(FILE *f, int close_handle_on_cl
 #ifdef STB_VORBIS_SDL
 extern stb_vorbis * stb_vorbis_open_rwops_section(SDL_RWops *rwops, int close_on_free, int *error, const stb_vorbis_alloc *alloc, unsigned int length);
 extern stb_vorbis * stb_vorbis_open_rwops(SDL_RWops *rwops, int close_on_free, int *error, const stb_vorbis_alloc *alloc);
+#define RWOPS_BUFFER_SIZE 2048
 #endif
 
 extern int stb_vorbis_seek_frame(stb_vorbis *f, unsigned int sample_number);
@@ -834,6 +835,10 @@ struct stb_vorbis
 #ifdef STB_VORBIS_SDL
    SDL_RWops *rwops;
    uint32 rwops_start;
+   uint32 rwops_virtual_pos;
+   uint32 rwops_buffer_pos;
+   uint32 rwops_buffer_fill;
+   uint8 rwops_buffer[RWOPS_BUFFER_SIZE];
    int close_on_free;
 #endif
 
@@ -1398,9 +1403,13 @@ static int STBV_CDECL point_compare(const void *p, const void *q)
 static uint8 get8(vorb *z)
 {
    #ifdef STB_VORBIS_SDL
-   uint8 c;
-   if (SDL_RWread(z->rwops, &c, 1, 1) != 1) { z->eof = TRUE; return 0; }
-   return c;
+   if (z->rwops_buffer_pos >= z->rwops_buffer_fill) {
+      z->rwops_buffer_fill = SDL_RWread(z->rwops, z->rwops_buffer, 1, RWOPS_BUFFER_SIZE);
+      z->rwops_buffer_pos = 0;
+      if (z->rwops_buffer_fill == 0) { z->eof = TRUE; return 0; }
+   }
+   z->rwops_virtual_pos++;
+   return z->rwops_buffer[z->rwops_buffer_pos++];
 
    #else
    if (USE_MEMORY(z)) {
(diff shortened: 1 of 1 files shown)


bjorn:
My application isn't updated to SDL3 so I've done the SDL2 patch first, but I can look at adjusting it to SDL3 later.

sezero:
For C90 compilers:

```diff
--- stb_vorbis.h~
+++ stb_vorbis.h
@@ -1441,6 +1441,8 @@ static int getn(vorb *z, uint8 *data, int n)
 {
    #ifdef STB_VORBIS_SDL
    while (n > 0) {
+      int chunk;
+
       if (z->rwops_buffer_pos >= z->rwops_buffer_fill) {
          z->rwops_buffer_fill = SDL_RWread(z->rwops, z->rwops_buffer, 1, RWOPS_BUFFER_SIZE);
          z->rwops_buffer_pos = 0;
@@ -1450,7 +1452,7 @@ static int getn(vorb *z, uint8 *data, int n)
          }
       }
 
-      int chunk = z->rwops_buffer_fill - z->rwops_buffer_pos;
+      chunk = z->rwops_buffer_fill - z->rwops_buffer_pos;
       if (chunk > n) chunk = n;
 
       memcpy(data, z->rwops_buffer + z->rwops_buffer_pos, chunk);
@@ -1510,6 +1512,8 @@ static int set_file_offset(stb_vorbis *f, unsigned int loc)
    f->eof = 0;
 
    #ifdef STB_VORBIS_SDL
+ {
+   unsigned int rwops_pos;
    uint32 buffer_start = f->rwops_virtual_pos - f->rwops_buffer_pos;
    uint32 buffer_end = buffer_start + f->rwops_buffer_fill;
    f->rwops_virtual_pos = loc;
@@ -1521,7 +1525,7 @@ static int set_file_offset(stb_vorbis *f, unsigned int loc)
       return 1;
    }
 
-   unsigned int rwops_pos = loc + f->rwops_start;
+   rwops_pos = loc + f->rwops_start;
    if (rwops_pos < loc || loc >= 0x80000000) {
       rwops_pos = 0x7fffffff;
       f->eof = 1;
@@ -1533,6 +1537,7 @@ static int set_file_offset(stb_vorbis *f, unsigned int loc)
    f->eof = 1;
    SDL_RWseek(f->rwops, f->rwops_start, RW_SEEK_END);
    return 0;
+ }
 
    #else
    if (USE_MEMORY(f)) {
```

slouken:
This looks good! Can you apply @sezero's changes, and I'll go ahead and merge this?

icculus:
(This is worth fixing in SDL_mixer, but to be clear: PhysicsFS lets you buffer data, too: PHYSFS_setBuffer())

sezero:
I took the libert of amending the C90 patch to this branch: OK to merge?

And, here is the SDL3 version of the patch: [671-3.patch](https://github.com/user-attachments/files/18933076/671-3.patch)
OK to apply after this? (Or can create a PR..)

bjorn:
> This is worth fixing in SDL_mixer, but to be clear: PhysicsFS lets you buffer data, too: PHYSFS_setBuffer())

Great, that would have saved quite some time! A quick comparison for reading a particular OGG file with a 2048 byte buffer yielded:

* No buffer: 47.59 ms
* Using `SDL_RWops` wrapping buffer: 0.86 ms
* Using `PHYSFS_setBuffer`: 1.41 ms
* Using buffer in stb_vorbis: 0.65 ms

So `PHYSFS_setBuffer` is indeed an acceptable solution even if it has a bit more overhead. This patch is still preferred, even if it's now mostly just because it doesn't require the user to set up the buffer explicitly.

@sezero Thanks for the C90 fixes and for porting this patch to SDL3! I would only suggest to amend the commit message to note that you've ported it over, but otherwise feel free to apply it wherever it helps.

(btw, if SDL_mixer is supposed to compile as C90, maybe the CMake project could be set up to enforce this?)

sezero:
OK, patches are in. Applied to SDL_sound, too.

Thanks.

bjorn:
> (At @bjorn's request, the SDL3 version should have you as the author, and just note it was based on his work in the commit message, though.)

Hmm, actually when I said "note that you've ported it over" I meant the other way around. I'm a bit sad that this way I don't show up as contributor on GitHub.

I'm glad the patch is in, though!

Issue #672, opened 2025-03-07
Incorporate SDL_sound


icculus:
In the SDL_mixer rewrite, Sam and I are toying with the idea of using SDL_sound for audio format decoding. This makes the mixer piece pretty light because a ton of code is going to evaporate from it in this case, and also offers an SDL_image-like equivalent library for audio files that can be used directly if the app doesn't care about the mixing part.

The TODO on this would be:

- [ ] Decide if this is the right thing to do in the first place.
- [ ] Migrate icculus/SDL_sound to libsdl-org.
- [ ] Give SDL_sound the full GitHub Actions treatment.
- [ ] Decide which decoders are in both SDL_sound and SDL_mixer, pick and choose what's best.
- [ ] Look over the SDL_sound API to bring it more in line with SDL3 conventions.
- [ ] Reformat SDL_sound's source code to be SDL coding style instead of Ryan-from-the-1990's style.

SDL_sound 2.0 [had a policy](https://www.patreon.com/posts/20211701) that all decoders needed to either be written from scratch or be (preferrably single-header) public domain code we could compile into the library directly. If we're backing SDL_mixer, this policy can be loosened if appropriate (for example, it might make sense to dump stb_vorbis for xiph.org's libvorbis, which can still be compiled in trivially, etc).

CC @madebr and @sezero for input, since this is likely going to cause work for both of you, so definitely tell me not to do this if it's a terrible idea.

icculus:
Formats in SDL_mixer that aren't in SDL_sound:

- "sndfile" uses libsndfile, which ostensibly supports [these formats](https://libsndfile.github.io/libsndfile/formats.html)...there's a lot of overlap in here and formats we probably don't care about. We might dump this.
- "native midi", use OS facilities for MIDI playback, external to SDL_mixer's mixing. Would like to dump this.
- "gme": chiptunes, played through libgme.
- "fluidsynth": midi decoding through FluidSynth
- "opus": Opus decoding through libopus. We'd take this into SDL_sound, which doesn't have this currently because of the no-external-library policy, but libopus can be compiled in and is...BSD licensed (?).
- 17 different things that play mod/tracker files. SDL_sound has modplug at the moment. Don't know which of these are good and bad.
- "wavpack": this doesn't use an external library, we'd pull it into SDL_sound.


[...]

PR #673, opened 2025-03-09
remove libmodplug support.

sezero:
As discussed in https://github.com/libsdl-org/SDL_mixer/issues/672

@icculus, @slouken: What to do with `MUS_MODPLUG_UNUSED` enumerated value
in the public interface? Just remove it? (We have no SDL3_mixer releases
anyway. Same for the `MUS_MP3_MAD_UNUSED` thing.)

P.S.: Should I do the same in SDL2 branch, or keep it as is?


sezero:
(CC @madebr for cmake side, just in case)

madebr (approved):

slouken:
This seems fine to me. I'll let @icculus merge when he's ready, I don't know what local changes he has at the moment.

We should leave SDL2 alone.

slouken:
> OK, I went ahead and removed the obsolete enum values MUS_MODPLUG_UNUSED and MUS_MP3_MAD_UNUSED.
> 
> @icculus: merge?

Yes, go ahead and merge.

sezero:
> Yes, go ahead and merge.

Done

PR #679, opened 2025-03-17
native_midi_linux_alsa: Add support for System Exclusive messages

tatokis:
This PR implements support for sending SysEx messages to MIDI devices using the native MIDI ALSA implementation.

Tested by playing back various SMF files containing SysEx messages and comparing them to aplaymidi using aseqdump.

```
$ ALSA_OUTPUT_PORTS=aseqdump ./playmus file.mid
$ ALSA_OUTPUT_PORTS=aseqdump aplaymidi file.mid
```

### DIFF
--- a/src/codecs/native_midi/native_midi_linux_alsa.c
+++ b/src/codecs/native_midi/native_midi_linux_alsa.c
@@ -42,6 +42,8 @@
 #include <string.h>
 #include <sys/stat.h>
 
+#include <assert.h>
+
 //#define SDL_NATIVE_MIDI_ALSA_DYNAMIC "libasound.so.2"
 
 static int load_alsa_syms(void);
@@ -438,7 +440,7 @@ static NativeMidiSong *currentsong = NULL;
 NativeMidiSong *native_midi_loadsong_IO(SDL_IOStream *src, bool closeio)
 {
     NativeMidiSong *song;
-    MIDIEvent *end;
+    MIDIEvent *event;
     int sv[2];
 
     if (!(song = SDL_calloc(1, sizeof(NativeMidiSong)))) {
@@ -455,7 +457,7 @@ NativeMidiSong *native_midi_loadsong_IO(SDL_IOStream *src, bool closeio)
     song->mainsock = sv[0];
     song->threadsock = sv[1];
 
-    end = song->evtlist = CreateMIDIEventList(src, &song->ppqn);
+    event = song->evtlist = CreateMIDIEventList(src, &song->ppqn);
 
     if (!song->evtlist) {
         close_sockpair(song);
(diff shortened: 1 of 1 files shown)


sezero:
CC @fragglet

tatokis:
@sezero I just saw 62bb7bbef52ab27957751f586d03ee9961a4828f and I'm curious; which version of gcc had trouble with it? I'd expect even super old gcc to support gnu99/c99.

I'm asking to avoid having this PR also break something by accident.

sezero:
> @sezero I just saw [62bb7bb](https://github.com/libsdl-org/SDL_mixer/commit/62bb7bbef52ab27957751f586d03ee9961a4828f) and I'm curious; which version of gcc had trouble with it? I'd expect even super old gcc to support gnu99/partial c99.
> 
> I'm asking to avoid having this PR also break something by accident.

It was my old gcc4.4 on my very outdated centos 6.10. It still supports the feature, but when in c99 mode which, who knows why, isn't requested by the cmake'ry yet.

src/codecs/native_midi/native_midi_linux_alsa.c
```diff
@@ -464,6 +466,37 @@ NativeMidiSong *native_midi_loadsong_IO(SDL_IOStream *src, bool closeio)
...
     }
 
+    /* Since ALSA requires the starting F0 for SysEx, but MIDIEvent.extraData doesn't contain it, we must preprocess the list */
+    /* In addition, since we're going through the list, store the last event's time for looping purposes */
+    do {
+        /* Is this a SysEx? */
+        if (event->status == MIDI_CMD_COMMON_SYSEX && event->extraLen) {
+            /* Sanity check in case something changes in the future */
+            /* This is safe to do since we can't have an F0 manufacturer ID */
+            assert(event->extraData[0] != 0xF0);
```
sezero, line 476: Should we use SDL_assert or one of its variants here? (CC @slouken )

tatokis: For what it's worth, I used assert explicitly so that it gets compiled out in non debug builds.

IIRC SDL_assert makes it into more than just debug, and a normal user is never going to hit. It's only for if there's a change in SDL_mixer itself.

tatokis:
> > @sezero I just saw [62bb7bb](https://github.com/libsdl-org/SDL_mixer/commit/62bb7bbef52ab27957751f586d03ee9961a4828f) and I'm curious; which version of gcc had trouble with it? I'd expect even super old gcc to support gnu99/partial c99.
> > I'm asking to avoid having this PR also break something by accident.
> 
> It was my old gcc4.4 on my very outdated centos 6.10. It still supports the feature, but when in c99 mode which, who knows why, isn't requested by the cmake'ry yet.

Ah, understood. Might be worth looking into `set_target_properties(target PROPERTIES C_STANDARD 99)`, which should set gnu99 by default, at least on Linux with gcc/clang. (I do not remember what happens on other platforms.)

slouken (approved):

sezero:
@fragglet :  Is this OK with you too?

tatokis:
Rebased and applied the following for consistency.

```diff
diff --git a/src/codecs/native_midi/native_midi_linux_alsa.c b/src/codecs/native_midi/native_midi_linux_alsa.c
index 795a469d..21983817 100644
--- a/src/codecs/native_midi/native_midi_linux_alsa.c
+++ b/src/codecs/native_midi/native_midi_linux_alsa.c
@@ -473,7 +473,7 @@ NativeMidiSong *native_midi_loadsong_IO(SDL_IOStream *src, bool closeio)
         if (event->status == MIDI_CMD_COMMON_SYSEX && event->extraLen) {
             /* Sanity check in case something changes in the future */
             /* This is safe to do since we can't have an F0 manufacturer ID */
-            assert(event->extraData[0] != 0xF0);
+            assert(event->extraData[0] != MIDI_CMD_COMMON_SYSEX);
 
             /* Resize by + 1 */
             Uint8 *newData = SDL_realloc(event->extraData, event->extraLen + 1);
@@ -489,7 +489,7 @@ NativeMidiSong *native_midi_loadsong_IO(SDL_IOStream *src, bool closeio)
             /* Prepend the F0 */
             event->extraData = newData;
             SDL_memmove(event->extraData + 1, event->extraData, event->extraLen);
-            event->extraData[0] = 0xF0;
+            event->extraData[0] = MIDI_CMD_COMMON_SYSEX;
             event->extraLen++;
         }
```

sezero:
This is in. Thanks.

Issue #662, opened 2025-01-30
The SDL_Mixer API Redesign Megathread


icculus:
Here's my first pitch at a redesigned SDL_mixer API. I took the same approach as SDL_net, where all of it goes in the trash and we start from scratch.

The idea is this:
- You no longer have a separation between "chunks" and "music".
- Channels are replaced by "Mix_Source" objects. They are dynamic, you don't have a set block of channels.
- You load sound data into "Mix_Audio" objects, which can be used on multiple Mix_Sources at once.
- Most of the functionality of the existing library is still here, possibly in a better form. Gone is the "effects" API, because all of that code sucked, and was generally limited to stereo audio anyhow. In its place, there are hooks to edit audio as it is about to be mixed, where any effect could be implemented, and the ability to wire up an SDL_AudioStream to a Mix_Source, so you can put any dynamic audio into the mixer that you want, as just another source to be mixed. The post-mix callback remains, too, which can be used for what the current library calls "posteffects."

This API spec is _extremely_ terse, so don't be afraid to ask questions. Obviously nothing is locked down at this point, so feedback is definitely welcome.

```c
// there is no separate "init" function. You open the audio device (presumably just the default audio device) and go.

bool Mix_OpenAudio(SDL_AudioDeviceID devid);  // this will call SDL_Init(SDL_INIT_AUDIO), open audio device.
void Mix_CloseAudio(void);  // this will call SDL_QuitSubSystem(SDL_INIT_AUDIO).


int Mix_GetNumAudioDecoders(void);
const char *Mix_GetAudioDecoder(int index);  // "WAV", "MP3", etc.

bool Mix_QuerySpec(SDL_AudioSpec *spec);   // what the device is actually expecting.

// there is no difference between sounds and "music" now. They're all Mix_Audio objects.
Mix_Audio *Mix_LoadAudio_IO(SDL_IOStream *src, bool closeio);
Mix_Audio *Mix_LoadAudio(const char *path);
Mix_Audio *Mix_LoadAudioWithProperties(SDL_PropertiesID props);  // lets you specify things like "here's a path to MIDI instrument files outside of this file", etc.

SDL_PropertiesID Mix_GetAudioProperties(Mix_Audio *audio);  // we can store audio format-specific metadata in here (artist/album/etc info...)

void Mix_DestroyAudio(Mix_Audio *audio);  // reference-counted; if this is playing, it will be _actually_ destroyed when no longer in use.


[...]

PR #689, opened 2025-04-03
Add `userdata` argument to `Mix_ChannelFinished` callback

### LINKED ISSUE #161: make Mix_ChannelFinished work in C++
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** unspecified
**Reported for operating system, platform:** All, All

# Comments on the original bug report:

On 2014-03-23 13:38:43 +0000, Daniel Sobe wrote:

> Hi,
> 
> I suggest to change the Mix_ChannelFinished callback (or add a new appropriate one if API compatibility is a must). The patch below changes the callback to also carry a void* argument, so that C++ objects can be supplied, and thus C++ can utilize the callback.
> 
> --- SDL_mixer/SDL_mixer.h	2014-03-23 14:04:43.089323652 +0100
> +++ SDL_mixer_new/SDL_mixer.h	2014-03-23 11:04:40.000000000 +0100
> @@ -223,7 +223,7 @@
>   *  inside the audio callback, or SDL_mixer will explicitly lock the audio
>   *  before calling your callback.
>   */
> -extern DECLSPEC void SDLCALL Mix_ChannelFinished(void (*channel_finished)(int channel));
> +extern DECLSPEC void SDLCALL Mix_ChannelFinished(void *userdata, void (*channel_finished)(void *userdata, int channel));
>  
>  
>  /* Special Effects API by ryan c. gordon. (icculus@icculus.org) */
> 
> --- SDL_mixer/mixer.c	2014-03-23 14:04:46.421417431 +0100
> +++ SDL_mixer_new/mixer.c	2014-03-23 11:17:49.000000000 +0100
> @@ -93,7 +94,8 @@
>  static void *mix_postmix_data = NULL;
>  
>  /* rcg07062001 callback to alert when channels are done playing. */
> -static void (*channel_done_callback)(int channel) = NULL;
> +static void (*channel_done_callback)(void *userdata, int channel) = NULL;
> +static void *channel_done_callback_userdata;
>  
>  /* Music function declarations */
>  extern int open_music(SDL_AudioSpec *mixer);
> @@ -261,7 +317,7 @@
>  static void _Mix_channel_done_playing(int channel)
>  {
>      if (channel_done_callback) {
> -        channel_done_callback(channel);
> +        channel_done_callback(channel_done_callback_userdata, channel);
>      }
>  
>      /*
> @@ -842,10 +903,11 @@
>      return(music_data);
>  }
>  
> -void Mix_ChannelFinished(void (*channel_finished)(int channel))
> +void Mix_ChannelFinished(void *userdata, void (*channel_finished)(void *userdata, int channel))
>  {
>      SDL_LockAudio();
>      channel_done_callback = channel_finished;
> +    channel_done_callback_userdata = userdata;
>      SDL_UnlockAudio();
>  }

On 2017-10-21 20:07:05 +0000, Sam Lantinga wrote:

> This is a good idea, but needs to wait for a breaking ABI rev

On 2017-10-28 20:36:55 +0000, Ivan Epifanov wrote:

> Having userdata in callback would be great not only from c++ POV.
> But, you can use this in c++ even now by utilizing std::bind and some more magic. See https://github.com/falltergeist/falltergeist/blob/develop/src/Audio/Mixer.cpp#L154 for example.

linkdd:
Close #161

### DIFF
--- a/include/SDL3_mixer/SDL_mixer.h
+++ b/include/SDL3_mixer/SDL_mixer.h
@@ -1052,7 +1052,7 @@ extern SDL_DECLSPEC void SDLCALL Mix_HookMusicFinished(Mix_MusicFinishedCallback
  */
 extern SDL_DECLSPEC void * SDLCALL Mix_GetMusicHookData(void);
 
-typedef void (SDLCALL *Mix_ChannelFinishedCallback)(int channel);
+typedef void (SDLCALL *Mix_ChannelFinishedCallback)(void *userdata, int channel);
 
 /**
  * Set a callback that runs when a channel has finished playing.
@@ -1065,12 +1065,13 @@ typedef void (SDLCALL *Mix_ChannelFinishedCallback)(int channel);
  *
  * A NULL pointer will disable the callback.
  *
+ * \param userdata the user data to pass to the callback.
  * \param channel_finished the callback function to become the new
  *                         notification mechanism.
  *
  * \since This function is available since SDL_mixer 3.0.0.
  */
-extern SDL_DECLSPEC void SDLCALL Mix_ChannelFinished(Mix_ChannelFinishedCallback channel_finished);
+extern SDL_DECLSPEC void SDLCALL Mix_ChannelFinished(void *userdata, Mix_ChannelFinishedCallback channel_finished);
 
 /**
  * Magic number for effects to operate on the postmix instead of a channel.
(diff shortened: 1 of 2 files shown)


icculus:
This breaks ABI on the existing library, so I probably wouldn't merge this.

(In the in-progress rewrite, this has a userdata argument already.)

linkdd:
I thought the major release 3.0 was the right moment to break ABI.

Where is the in-progress rewrite btw? I wanted to help because I use SDL3 and wished for SDL3_mixer, but I see no clear communication on the subject.

icculus:
> Where is the in-progress rewrite btw?

The redesign is being discussed over in #662, and the work-in-progress is temporarily at https://github.com/icculus/SDL_remixer. Definitely check it out and give feedback, all current decisions are still able to change!

PR #692, opened 2025-04-08
cmake: add current directory to module search path in installed config file

madebr:


sezero:
Looks like this solves the 'Linux (CMake, static)' workflow failing
at 'verify cmake configuration files' step, because it couldn't find
PkgConfigHelper.cmake: you're still not installing those cmake stuff
in static builds but it now succeeds?

In any case, is this relevant in other satellite libs?

madebr:
> Looks like this solves the 'Linux (CMake, static)' workflow failing at 'verify cmake configuration files' step, because it couldn't find PkgConfigHelper.cmake: you're still not installing those cmake stuff in static builds but it now succeeds?

We were installing it, but we are including it without an absolute path, so we need to add the current directory to the module path.

We have to modify the module search path anyways, otherwise the various `find_package`'s would fail finding our custom `FindXXX.cmake` modules.


> 
> In any case, is this relevant in other satellite libs?

I did a search and we correctly modify `CMAKE_MODULE_PATH` in other SDL2 projects.

sezero (approved):

sezero:
OK then, it's in.

PR #698, opened 2025-05-16
mpg123: Use "portable API" from v1.32+ if available

smcv:
The libmpg123 ABI has been through several incompatible versions, and depends on the size of off_t, which can vary with compiler options, and can therefore cause build or load failures if SDL_mixer and libmpg123 are built with different values for _FILE_OFFSET_BITS.

In particular, the latest snapshots of SDL3_mixer are failing to link as packaged in Debian, with libmpg123 configured as a hard dependency, on 32-bit architectures whose compiler defaults to a 64-bit off_t. Debian carried out an ABI transition from 32-bit to 64-bit off_t on 32-bit ARM and 32-bit -ports architectures (but not on 32-bit x86!) during early 2024.

SDL itself avoids this problem by always using 64-bit offsets (Sint64) in its own API/ABI.

Recent versions of libmpg123 similarly have an API subset that consistently uses 64-bit offsets (Standard C int64_t, equivalent to SDL's Sint64). Use that API, where available.

---

I think this is ready for review, but I'll leave it marked as draft until I've been able to try building this on one of the affected architectures.

### DIFF
--- a/src/codecs/music_mpg123.c
+++ b/src/codecs/music_mpg123.c
@@ -30,6 +30,7 @@
 #include "mp3utils.h"
 
 #include <stdio.h>      /* For SEEK_SET */
+#define MPG123_PORTABLE_API /* hide off_t-dependent APIs entirely if possible */
 #define MPG123_NO_LARGENAME /* disable the _FILE_OFFSET_BITS suffixes. */
 #ifdef MPG123_HEADER
 #include MPG123_HEADER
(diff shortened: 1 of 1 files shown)


smcv:
This isn't quite right, `mpg123_open_handle` also isn't found.

smcv:
I'll open an issue report rather than reverse-engineering what the issue is from the proposed solution.

I think using the "portable API" where available is probably desirable, but we might well need changes beyond that, if the goal is for SDL3_mixer to be 32-bit-compatible.

PR #700, opened 2025-05-23
cmake: static gme depends on gme_deps, so install that target

madebr:


sezero:
Is the same needed in SDL2 and release-2.8.x ?

PR #713, opened 2025-07-26
SDL_mixer_metadata_tags.c:parse_id3v1_ansi_string(): Fixed "always true" expression

Sackzement:
The comparison `i >= 0` is always true because `i` is an unsigned type.

```c
[  6%] Building C object CMakeFiles/SDL3_mixer-shared.dir/src/SDL_mixer_metadata_tags.c.o
/path/to/SDL_mixer/src/SDL_mixer_metadata_tags.c: In function ‘parse_id3v1_ansi_string’:
/path/to/SDL_mixer/src/SDL_mixer_metadata_tags.c:99:37: warning: comparison of unsigned expression in ‘>= 0’ is always true [-Wtype-limits]
   99 |     for (size_t i = src_len - 1; (i >= 0) && (src_buffer[i] == ' '); i--) {
      |                                     ^~
```

This commit uses a do-while loop instead.

_Might be a bit more "chatty". :)_
_4 lines vs 12 lines_

### DIFF
--- a/src/SDL_mixer_metadata_tags.c
+++ b/src/SDL_mixer_metadata_tags.c
@@ -96,9 +96,17 @@ static char *parse_id3v1_ansi_string(const Uint8 *buffer, size_t src_len)
     src_buffer[src_len] = '\0';
 
     // trim whitespace from end (some id3v1 tags pad out with space instead of nulls).
-    for (size_t i = src_len - 1; (i >= 0) && (src_buffer[i] == ' '); i--) {
-        src_buffer[i] = '\0';
-        src_len--;
+    if (src_len > 0) {
+        size_t i = src_len;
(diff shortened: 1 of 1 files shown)


Sackzement:
Totally forgot the `break;`for the second condition. It's fixed now.

slouken:
Merged, thanks!

sezero:
And _why_ is the loop iterator in there is a size_t instead of a plain signed int??

Sackzement:
> And _why_ is the loop iterator in there is a size_t instead of a plain signed int??

`src_len` is also `size_t`. The loop iterator just uses the same type as what is passed to the function.
My guess is because the buffer can be very large which `int` wouldn't be able to index.

Issue #711, opened 2025-07-25
Use type `Sint64` for frames/milliseconds


Sackzement:
The function `MIX_SetTrackPlaybackPosition()` and the frame/millisecond conversion functions are currently the only ones that use `Uint64`
for frames/milliseconds parameters.

I would like to propose to change the types to `Sint64`.

For the function:
```c
bool MIX_SetTrackPlaybackPosition(MIX_Track *track, Uint64 frames)
```
... it just doesn't make sense to have a parameter type of `Uint64` because no other function can get or set a frame/millisecond value with the type `Uint64`.

Full list of functions that use frames/milliseconds and their types:
```c
       | SDL_PropertiesID MIX_GetAudioProperties(MIX_Audio *audio)
Sint64 | Sint64 SDL_GetNumberProperty(SDL_PropertiesID props, const char *name, Sint64 default_value)
       | MIX_PROP_METADATA_DURATION_FRAMES_NUMBER
       | 
Sint64 | Sint64 MIX_GetAudioDuration(MIX_Audio *audio) // -1 -> MIX_DURATION_UNKNOWN, -2 -> MIX_DURATION_INFINITE
       | 
Uint64 | bool   MIX_SetTrackPlaybackPosition(MIX_Track *track, Uint64 frames)
Sint64 | Sint64 MIX_GetTrackPlaybackPosition(MIX_Track *track) // -1 on error
       | 
Sint64 | Sint64 MIX_GetTrackRemaining(MIX_Track *track) // -1 if unknown
       | 
Uint64 | Uint64 MIX_TrackMSToFrames(MIX_Track *track, Uint64 ms    ) // returns 0 on error, but could be valid
Uint64 | Uint64 MIX_TrackFramesToMS(MIX_Track *track, Uint64 frames) // returns 0 on error, but could be valid
Uint64 | Uint64 MIX_AudioMSToFrames(MIX_Audio *audio, Uint64 ms    ) // returns 0 on error, but could be valid
Uint64 | Uint64 MIX_AudioFramesToMS(MIX_Audio *audio, Uint64 frames) // returns 0 on error, but could be valid
Uint64 | Uint64 MIX_MSToFrames     (int sample_rate , Uint64 ms    ) // returns 0 on error, but could be valid
Uint64 | Uint64 MIX_FramesToMS     (int sample_rate , Uint64 frames) // returns 0 on error, but could be valid
       | 
[...]

PR #718, opened 2025-07-28
Changed frames/ms types from Uint64 to Sint64

Sackzement:
This is the first draft for converting the parameter types of of `frames`/`ms` for the conversion functions and the `MIX_SetTrackPlaybackPosition()` function from `Uint64` to `Sint64`.

Conversion functions now return `-1` on error and set a message.
Negative `frame`/`ms` arguments are not allowed.

TODO:
Comments could use a little more tweaking.

Maybe?:
Allow converting negative `frames`/`ms` so that negative return values can be used in arithmetic expressions? (return value on error would then be 0 again)

Closes https://github.com/libsdl-org/SDL_mixer/issues/711


icculus:
> Allow converting negative frames/ms so that negative return values can be used in arithmetic expressions?

I was thinking about this, too...but let's merge this for now and we can consider that separately.

PR #722, opened 2025-07-29
Make `mixer` optional in `MIX_LoadRawAudio_IO`

alexbatalov:
The docs state it's optional as in other `MIX_Load*` functions.


icculus:
Good catch, thanks!

Issue #719, opened 2025-07-28
MIX_LoadAudio segfault


1vanK:
If PC have no audio card (Github actions server for example) and
```
mixer_ = MIX_CreateMixerDevice(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, nullptr);
```
returns NULL with error message `ALSA: Couldn't open audio device: No such file or directory`

`MIX_LoadAudio(mixer_, ...);` cause SegFault

icculus:
I can't reproduce this, even back to the SDL_remixer merge commit.

Modified SDL3 to unconditionally fail to open a PipeWire device:

```diff
diff --git a/src/audio/pipewire/SDL_pipewire.c b/src/audio/pipewire/SDL_pipewire.c
index a9e95a6fe..19510f480 100644
--- a/src/audio/pipewire/SDL_pipewire.c
+++ b/src/audio/pipewire/SDL_pipewire.c
@@ -1077,6 +1077,7 @@ static const struct pw_stream_events stream_input_events = { PW_VERSION_STREAM_E
 
 static bool PIPEWIRE_OpenDevice(SDL_AudioDevice *device)
 {
+return SDL_SetError("Forced to fail.");
     /*
      * NOTE: The PW_STREAM_FLAG_RT_PROCESS flag can be set to call the stream
      * processing callback from the realtime thread.  However, it comes with some
```

Then used this test program:

```c
#include <SDL3/SDL.h>
#include <SDL3_mixer/SDL_mixer.h>
#include <SDL3/SDL_main.h>

int main(int argc, char **argv)
{
    MIX_Init();

    MIX_Mixer *mixer_ = MIX_CreateMixerDevice(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, NULL);
    if (!mixer_) {
        SDL_Log("MIX_CreateMixerDevice() failed: %s", SDL_GetError());
    }
    MIX_Audio *audio = MIX_LoadAudio(mixer_, "click.flac", false);
    if (!audio) {
        SDL_Log("MIX_LoadAudio() failed: %s", SDL_GetError());
    }

    MIX_Quit();

    return 0;
}
```

This gives me...

```
MIX_CreateMixerDevice() failed: Forced to fail.
```

...but the MIX_LoadAudio() call still gets made with a NULL `mixer_`, and works without crashing.


I'm going to need a better reproduction case for this, if it's still happening to you.

1vanK:
[...]

PR #723, opened 2025-07-29
Validate `mixer` in `MIX_PlayAudio`

alexbatalov:
Causes segfault if `NULL` is being passed. Probably refers to #719, not sure.


slouken:
Merged, thanks!

Issue #733, opened 2025-08-03
build error on android with NEON intrinsics


1bsyl:
Hi,  latest version fails to compile on Android with ndk r27c
 
```
src/SDL_mixer_spatialization.c:320:31: error: call to undeclared function 'vcopyq_laneq_f32'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  320 |     const float32x4_t a_yzx = vcopyq_laneq_f32(vextq_f32(a, a, 1), 2, a, 0);
      |                               ^
src/SDL_mixer_spatialization.c:320:23: error: initializing 'const float32x4_t' (vector of 4 'float32_t' values) with an expression of incompatible type 'int'
  320 |     const float32x4_t a_yzx = vcopyq_laneq_f32(vextq_f32(a, a, 1), 2, a, 0);
      |                       ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/lib/clang/18/include/arm_neon.h 
seems to define this function as a macros ... 

same error, if I add` #include <arm_neon.h>` in SDL_mixer_spatialization.c

Any idea ? 
thanks

icculus:
Yeah, we changed this to make Visual Studio happy; I might just put this in if/else preprocessor blocks for now.

1bsyl:
In the android ndk, arm_neon.h,  "vcopyq_laneq_f32" is a macro,
but it is not activated because this condition is false:
`#if (__ARM_FP & 2)`
it appears it is defined as: `#define __ARM_FP 0xc`

Let's just unactivate this for now

1bsyl:
This is fixed in PR https://github.com/libsdl-org/SDL_mixer/pull/737

PR #737, opened 2025-08-04
Fixed bug #733: build error on android with NEON intrinsics

1bsyl:
Neon intrinsics "vcopyq_laneq_f32" is undeclared.
It is defined in <arm_neon.h>, but not activated because of __ARM_FP value.

So, unactivate NEON intrinsics for spatialization on Android.

### DIFF
--- a/src/SDL_mixer_spatialization.c
+++ b/src/SDL_mixer_spatialization.c
@@ -314,6 +314,12 @@ static void SDL_TARGETING("sse") calculate_distance_attenuation_and_angle_sse(co
 }
 #endif
 
+// FIXME: neon intrinsics "vcopyq_laneq_f32" is undeclared.
+// It is defined in <arm_neon.h>, but not activated because of __ARM_FP value.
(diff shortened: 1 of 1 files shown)


sezero:
Without this patch, does it still fail when targeting arm64 ??

1bsyl:
@sezero 
indeed, you're right
only armeabi-v7a is affected
arm64-v8a is fine

sezero:
So, add an additional restriction for arm64, e.g. something like this ???

```c
#if defined(SDL_PLATFORM_ANDROID) && defined(SDL_NEON_INTRINSICS) && !defined(__aarch64__)
```

1bsyl:
@sezero 
yes, that works! thanks.
I can commit that later

sezero:
> @sezero yes, that works! thanks.

You're welcome

> I can commit that later

Yes please.  Otherwise, it would be sad to disable neon for aarch64.

slouken:
Committed!

PR #754, opened 2025-09-04
cmake: fluidsynth package can contain versioned fluidsynth libraries

### LINKED ISSUE #755: MingW broken?
It looks like b0503d282ede2a6a91eb9e0f97a3d5c8e3147017 broke MingW on the builders...it looks like it's looking for FluidSynth 2.2.0, finds 2.4.7, and panics.

```
CMake Error at D:/a/_temp/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
  Could NOT find FluidSynth (missing: FluidSynth_LIBRARY) (found suitable
  version "2.4.7", minimum required is "2.2.0")
```

https://github.com/libsdl-org/SDL_mixer/actions/runs/17447021195/job/49543866718

madebr:
This was changed in https://github.com/FluidSynth/fluidsynth/pull/1546

Closes #755

### DIFF
--- a/cmake/FindFluidSynth.cmake
+++ b/cmake/FindFluidSynth.cmake
@@ -4,7 +4,7 @@ find_package(PkgConfig QUIET)
 pkg_check_modules(PC_FLUIDSYNTH QUIET fluidsynth)
 
 find_library(FluidSynth_LIBRARY
-    NAMES fluidsynth libfluidsynth
+    NAMES fluidsynth fluidsynth-3 libfluidsynth
     HINTS ${PC_FLUIDSYNTH_LIBDIR}
 )
 


sezero:
As far as I can see this is a cheat / a bandaid n our part to workaround mainstream's breakage, and a cmake-specific one at that? What does their *.pc file look like? Becuease if it says -lfludsynth, other build systems, autotools for e..g., are broken as well, and we should wait for mainstream to fix their sh^w^w thing instead and not do this?

madebr:
Looks like they require the suffix:
```pc
prefix=/mingw64
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: FluidSynth
Description: Software SoundFont synth
Version: 2.4.8
Requires.private: glib-2.0 gthread-2.0 sndfile portaudio-2.0 gmodule-2.0 sdl3 readline
Libs: -L${libdir} -lfluidsynth-3
Libs.private: -ldsound -lksuser -lwinmm -lole32 -lws2_32 ${prefix}/lib/libgomp.dll.a ${prefix}/lib/libmingwthrd.a
Cflags: -I${includedir}
```

sezero:
> Looks like they require the suffix:

Geez.. Where the hell did they get that idea...

sezero:
Mainstrean issue: https://github.com/FluidSynth/fluidsynth/issues/1645

> > Looks like they require the suffix:
> 
> Geez.. Where the hell did they get that idea...

Looks like it happened in https://github.com/FluidSynth/fluidsynth/pull/1548 as a followup
to https://github.com/FluidSynth/fluidsynth/pull/1546

sezero:
@madebr: Looks like wour patch proposal at fludsynth mainstream isn't going anywhere anytime soon, so I guess we apply this patch here and be done with it?

P.S.: The bad thing about the approach in this patch is that it's too dependent on mainstream's so version changes. Today it's 3, may be it'll be 4 tomorrow, or 14, or 24., or 124..

sezero (approved):
This should be OK for the immediate problem, and will need backporting to all other branches.

That it' not future-proof is an issue will need addressing later.

madebr:
My upstream PR has not been closed yet , so there's still hope :)

PR #756, opened 2025-09-05
Fix: missing set free_precache

brccabral:
Simple call to `MIX_LoadAudio(mixer, filepath.c_str(), false)` then `MIX_DestroyAudio` was skipping `SDL_free` in `UnrefAudio`


slouken:
Merged, thanks!

PR #767, opened 2025-09-21
BugFix: FIX MIX_CreateSineWaveAudio(). 

OvJat:
MIX_CreateSineWaveAudio() has a bug: MIX_LoadAudioWithProperties() require MIX_PROP_AUDIO_LOAD_IOSTREAM_POINTER, but MIX_CreateSineWaveAudio() not set it, MIX_LoadAudioWithProperties always fail.

I set a dummy SDL_IOStream to walk around it , and it works.

PS: The current Unit Test only cover a little code, shall we write high cover Unit Test using GoogleTest?


src/SDL_mixer.c
```diff
@@ -1262,7 +1262,13 @@ MIX_Audio *MIX_CreateSineWaveAudio(MIX_Mixer *mixer, int hz, float amplitude)
     if (!props) {
         return NULL;
     }
-
+    // BugFix: MIX_LoadAudioWithProperties() require MIX_PROP_AUDIO_LOAD_IOSTREAM_POINTER, so we set dummy SDL_IOStream.
```
sezero, line 1265: This comment lne here is not necessary, please remove it.

OvJat: ok, I have remove it.

icculus:
Let me see if we can reasonably make this work without making a dummy iostream before we merge this.

icculus:
Fixed in 8cc5d4e7667d0b7217af332cb02c867115d3283e without making a throwaway IOStream. Thanks for taking time to work on this PR, though!

PR #781, opened 2025-11-17
cmake+xcode: add exports file for Apple

madebr:
I'm not 100% sure about the xcode project since it differs a bit from that of SDL_image and SDL_ttf:
it has no:
```
        OTHER_LDFLAGS = "$(CONFIG_FRAMEWORK_LDFLAGS)";
```


slouken:
Yep, this works.

slouken (approved):

PR #791, opened 2025-12-11
Include <opusfile.h> instead of <opus/opusfile.h>

### LINKED ISSUE #790: Upstream opusfile package is incompatible
Upstream opusfile exports includepath /include/opus not /include, when not using vendored mode this leads to bugs, how should this be handled?

madebr:
Fixes #790

### DIFF
--- a/cmake/FindOpusFile.cmake
+++ b/cmake/FindOpusFile.cmake
@@ -9,7 +9,8 @@ find_library(OpusFile_LIBRARY
 )
 
 find_path(OpusFile_INCLUDE_PATH
-    NAMES opus/opusfile.h
+    NAMES opusfile.h
+    PATH_SUFFIXES opus
     HINTS ${PC_OPUSFILE_INCLUDEDIR}
 )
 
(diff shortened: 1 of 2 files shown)


slouken:
Does this work on system installed (non-vendored) opus installations?

madebr:
> Does this work on system installed (non-vendored) opus installations?

I think so (I always configure the SDL libraries with system libraries on my system).
Configuring with the following patch to force our custom `FindOpusFile.cmake` keeps working:
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,7 +382,7 @@ if(SDLMIXER_OPUS)
             list(APPEND PC_LIBS -l$<TARGET_FILE_BASE_NAME:opusfile> -l$<TARGET_FILE_BASE_NAME:opus> -l$<TARGET_FILE_BASE_NAME:ogg>)
         endif()
     else()
-        find_package(OpusFile ${required})
+        find_package(OpusFile MODULE ${required})
         if(OpusFile_FOUND AND Ogg_FOUND)
             set(SDLMIXER_OPUS_ENABLED TRUE)
             message(STATUS "Using system opusfile")
```

PR #803, opened 2026-01-20
Remap Vorbis channels to match SDL channels, and some `*UpdateSection()` changes.

dkosmari:
This adds a function to remap channels in Vorbis and Opus sources to SDL channels. Without it, front center/right get swapped, and all following channels get shuffled.

While looking at the surrounding code, I noticed that any change in the section (sample rate or number of channels) will discard the data that was just read, in the `*_UpdateSection()` functions. Instead of a free/malloc pair, I changed them to realloc the buffer (only if it's larger) in order to preserve the decoded data.

The `music_ogg_stb.c` backend had a bogus check for section change, so it would never really call `OGG_UpdateSection()`, if the number of channels or sample rate did change.

### DIFF
--- a/src/codecs/remap_channels.h
+++ b/src/codecs/remap_channels.h
@@ -28,4 +28,4 @@ extern void remap_channels_vorbis_s16(Sint16 *samples, int num_samples, int num_
 
 extern void remap_channels_vorbis_flt(float *samples, int num_samples, int num_channels);
 
-#endif /* CHANNEL_REMAP_H_ */
+#endif /* REMAP_CHANNELS_H_ */


sezero:
Seeing that this is for the SDL2 branch, I assume SDL3_mixer is good?

icculus:
> I assume SDL3_mixer is good?

Oh, I thought I had channel maps set up for this in SDL3_mixer, but apparently I don't. It definitely doesn't need _this_ patch--SDL3 audio streams can handle this for us--but I'll open a separate issue to deal with it.

sezero:
Is this OK for release-2.8.x branch too, or is it considered too much of a behavior change?

icculus:
I guess that's a question for @slouken: are we even doing another SDL2_mixer release?

I'm still fixing these things up in SDL3_mixer, but none of that is going to apply to the SDL2 branch.

slouken:
> I guess that's a question for @slouken: are we even doing another SDL2_mixer release?
> 
> I'm still fixing these things up in SDL3_mixer, but none of that is going to apply to the SDL2 branch.

We're not planning on it, though if things are merged to the SDL2 branch, they should probably be merged through to the next release branch, just in case.

sezero:
> We're not planning on it, though if things are merged to the SDL2 branch, they should probably be merged through to the next release branch, just in case.

OK, cherry picked to the release 2.8.x branch now

PR #810, opened 2026-01-22
Fix config typo

martinstarkov:
Small typo in release-info.json caused my CMake build to fail with:

```
Could not find a configuration file for package "SDL3_mixer" that is
  compatible with requested version "3.1.2".

  The following configuration files were considered but not accepted:

    C:/Dev/protegon/external/windows/msvc/SDL3_mixer-3.1.2/cmake/SDL3_mixerConfig.cmake, version: unknown
```


slouken:
Merged, thanks!

PR #823, opened 2026-02-28
change some `x == -1` checks into `x < 0` for simplicity

sezero:


slouken:
This is consistent with SDL code style, thanks!

PR #840, opened 2026-04-25
Update INTRO-visualstudio.md

tylawin:
Copied missing step from SDL_ttf instructions.

This is still incomplete instructions since it doesn't address timidity which is not in the git submodules.


slouken:
Merged, thanks!

Issue #857, opened 2026-05-19
Chrome: Dangerous download blocked


ivan-enzhaev:
- Chrome Version 148.0.7778.179 (Official Build) (64-bit)
- Windows 10

<img width="300" height="167" alt="Image" src="https://github.com/user-attachments/assets/b30e39f6-49ae-4302-8a57-ea63f1d6c6a6" />
<br>
---
<br>
<img width="306" height="210" alt="Image" src="https://github.com/user-attachments/assets/13fcd58a-961c-4c6d-a952-9d2fb5e7aa22" />

Firefox:

<img width="434" height="63" alt="Image" src="https://github.com/user-attachments/assets/b9bac128-bcc8-4c28-9b38-dafb5e67687e" />

madebr:
Duplicate of https://github.com/libsdl-org/SDL_mixer/issues/854?

ivan-enzhaev:
Closing as a duplicate of #854, thank you!

Issue #854, opened 2026-05-16
Bitdefender is detecting a trojan on 3.2.2 mingw release


Coquitos:
Bitdefender is  detecting a Trojan (Trojan.GenericKD.80075253) on https://github.com/libsdl-org/SDL_mixer/releases/download/release-3.2.2/SDL3_mixer-devel-3.2.2-mingw.zip and removing it
After restoring the file and extracting it, these are the specific files causing the problem.
SDL3_mixer-3.2.2\x86_64-w64-mingw32\libexec\installed-examples\SDL3_mixer\basics-load-and-play.exe is infected with Trojan,GenericKD.80128279
SDL3_mixer-3.2.2\x86_64-w64-mingw32\libexec\installed-examples\SDL3_mixer\basics-play-multiple-sounds.exe is infected with Trojan.GenericKD.80128288
SDL3_mixer-3.2.2\x86_64-w64-mingw32\libexec\installed-examples\SDL3_mixer\basics-play-with-options.exe is infected with Gen.Suspicious.Cloud.1.iBZ@a8AOkVI

Same files are tagged as infected in https://github.com/libsdl-org/SDL_mixer/releases/download/release-3.2.2/SDL3_mixer-devel-3.2.2-mingw.tar.gz
It may be a false positive. But I'm letting you know just in case.

Coquitos:
The same problem happens with 3.2.0 release

icculus:
Hmm, I don't love this (64-bit basics-play-with-options.exe from 3.2.2)

https://www.virustotal.com/gui/file/fd75eeefcab74279b1cc0a820c441f5460fe4ab5c94c08ecab73fb7cb7374b56

It could be a false-positive, I don't know. @slouken, where did we build these? Is it possible the system with the toolchain is infected, or if this is built on GitHub Actions, that the toolchain we're downloading to build this is infected?

slouken:
It's possible. We're building with GitHub actions. Maybe @madebr has more insight here?

madebr:
They are build using a mingw toolchain from ubuntu, using a static libc.
Perhaps it's because of the static runtime they may have in common?

icculus:
I'm fairly confident it's a false-positive, but ugh. I'm not even sure what we could do to change that.

madebr:
These scan sevices will only learn that SDL3_mixer is harmless if enough people mark it as harmless.

[...]

PR #858, opened 2026-05-22
cmake: add SDLMIXER_LIBC option allowing to disable libc usage shared lib

madebr:

By configuring with `-DSDLMIXER_LIBC=OFF`, you can build a SDL3_mixer shared library that does not use the c library.
I tested it using a MSVC and gcc linux toolchain.

It does not work for mingw at the moment because of a missing implementation for: `__chkstk_ms`, `__umoddi3`, `__udivdi3`, `__divmoddi4` for i686 and x86_64 (and maybe additional symbols)

### DIFF
--- a/VisualC/SDL_mixer.vcxproj
+++ b/VisualC/SDL_mixer.vcxproj
@@ -46,6 +46,7 @@
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
+    <Import Project="$(VCTargetsPath)\BuildCustomizations\masm.props" />
   </ImportGroup>
   <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="PropertySheets">
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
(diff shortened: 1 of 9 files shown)


sezero:
> It does not work for mingw at the moment because of a missing implementation for: `__chkstk_ms`, `__umoddi3`, `__udivdi3`, `__divmoddi4` for i686 and x86_64 (and maybe additional symbols)

`-static-libgcc -lgcc` for gcc-based mingw toolchains???

(clang-based ones will need appropriate -lcompiler_rt-xxx, I guess..)

madebr:
> `-static-libgcc -lgcc` for gcc-based mingw toolchains???
> 
> (clang-based ones will need appropriate -lcompiler_rt-xxx, I guess..)

No, we want no runtime at all, so also not as a static library.

sezero:
> No, we want no runtime at all, so also not as a static library.

They aren't libc runtimes, they are compiler builtins.

madebr:
> > No, we want no runtime at all, so also not as a static library.
> 
> They aren't libc runtimes, they are compiler builtins.

You're right. With your suggested changes, I can build mingw binaries.
I'm going to ignore static clang until somebody who needs it comes along and tells us the exact commands.

src/nolibc/SDL_mslibc.c
```diff
@@ -0,0 +1,841 @@
...
+
+#include <windows.h>
+
+#if defined(__GNUC__) && defined(__i386__) && defined(HAVE_FORCE_ALIGN_ARG_POINTER)
+#define MINGW32_FORCEALIGN __attribute__((force_align_arg_pointer))
+#else
+#define MINGW32_FORCEALIGN
+#endif
+
+#ifdef _MSVC_VER
```
sezero: `_MSC_VER`

(I guess this wasn't in SDL itself and/or sdl[1]2-compat ?)

madebr: Indeed. This was the only change needed to get the dll building using a mingw toolchain without libc.
I never built SDL3 using a mingw toolchain without libc.

I'm including it for #854 and #857

src/nolibc/SDL_mslibc.c
```diff
@@ -0,0 +1,841 @@
...
+{
+    return SDL_memset(dst, c, len);
+}
+#endif
+
+#ifdef SDL_PLATFORM_WIN32
+
+#include <windows.h>
+
+#if defined(__GNUC__) && defined(__i386__) && defined(HAVE_FORCE_ALIGN_ARG_POINTER)
```
sezero, line 789: I think `HAVE_FORCE_ALIGN_ARG_POINTER` and `MINGW32_FORCEALIGN` aren't defined anywhere in SDL_mixer.  https://github.com/libsdl-org/SDL/blob/main/src/core/windows/SDL_windows.h#L114-L125 should be copied here..

madebr: Indeed. Copied those lines. Luckily, the entry function simply returns TRUE, so stack alignment is not an  issue on optimized builds.

madebr:
Release builds without libc:
https://github.com/madebr/SDL_mixer/actions/runs/26309512923 (for mingw and msvc)

src/nolibc/SDL_mslibc.c
```diff
@@ -0,0 +1,854 @@
...
+#define MINGW32_FORCEALIGN __attribute__((force_align_arg_pointer))
+#else
+#define MINGW32_FORCEALIGN
+#endif
+
+#if defined(__GNUC__) && defined(__i386__) && defined(HAVE_FORCE_ALIGN_ARG_POINTER)
+#define MINGW32_FORCEALIGN __attribute__((force_align_arg_pointer))
+#else
+#define MINGW32_FORCEALIGN
+#endif
```
sezero: These are now duplicates of above lines 789-793 :)

madebr: Right, removed the duplicates now :)

madebr:
I added this to the release builds. I'm not sure we want this?

sezero:
> I'm not sure we want this?

Not 100% happy with it myself.

As for https://github.com/libsdl-org/SDL_mixer/issues/854 and https://github.com/libsdl-org/SDL_mixer/issues/857: I wish malware detectors would update their stuff and whitelist mingw code... My firefox still barfs, and even if I tell it to allow the download, then MS virus scanner gets in the way..

VisualC/SDL_mixer.vcxproj
```diff
@@ -116,7 +117,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(ProjectDir)..\include;$(ProjectDir)..\src;$(ProjectDir)..\external\flac\include;$(ProjectDir)..\external\libxmp\include;$(ProjectDir)..\external\ogg\include;$(ProjectDir)..\external\vorbis\include;$(ProjectDir)..\external\libgme;$(ProjectDir)\external\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>DLL_EXPORT;_DEBUG;WIN32;_WINDOWS;DECODER_WAV;DECODER_AIFF;DECODER_VOC;DECODER_AU;DECODER_WAVPACK;DECODER_FLAC_DRFLAC;DECODER_MOD_XMP;XMP_DYNAMIC="libxmp.dll";DECODER_MP3_DRMP3;DECODER_OGGVORBIS_VORBISFILE;DECODER_OGGVORBIS_STB;VORBIS_DYNAMIC="libvorbisfile-3.dll";DECODER_OPUS;OPUS_DYNAMIC="libopusfile-0.dll";WAVPACK_DYNAMIC="libwavpack-1.dll";DECODER_MIDI_TIMIDITY;DECODER_GME;GME_DYNAMIC="libgme.dll";%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
sezero, line 119: Because we provide neither headers nor dlls of ogg/vorbis libs under VisualC/external/?

If so, I guess these removals needs applying out of this P/R already?

madebr: Yes, these failed to build for me because of the missing headers.

sezero: OK, let's take those removals out of this P/R and apply them directly to main and to 3.2.x

slouken:
2 cents from the peanut gallery... this seems like a lot of work to avoid a false positive on mingw C runtime. 

We're not planning to disable the C runtime in general for release builds are we? 

The release "win32" builds are all built with Visual Studio, right?

sezero:
>  this seems like a lot of work to avoid a false positive on mingw C runtime.

Agreed.

madebr:
It was more of a proof of concept I guess.

PR #859, opened 2026-05-25
VisualC: vorbisfile headers are not present in VisualC's external directory

madebr:


sezero (approved):

icculus:
Should we just add the headers instead of disabling the vorbisfile backend?

sezero:
> Should we just add the headers instead of disabling the vorbisfile backend?

Headers _and_ dlls..

However, seeing how mp3 and flac are supported only by single-header options
and not by mpg123 and libflac, this patch is the way to go.

Otherwise, if we do add vorbis headers and dlls, we should do so for mpg123 and
libflac, too, for consistency.

slouken:
Yeah, it's not that this format isn't supported, it's that we're using stb_vorbis by default.

slouken (approved):

sezero:
Alright, merging this.

PR #866, opened 2026-06-13
ci: add yasm patch

madebr:
CMakeLists.txt from upstream yasm is not CMake 4.x compatible.

### DIFF
--- a/.github/actions/setup-yasm/0002-bool-is-a-C-keyword-from-C23-onwards.patch
+++ b/.github/actions/setup-yasm/0002-bool-is-a-C-keyword-from-C23-onwards.patch
@@ -1,0 +1,35 @@
+From f2dba873ec880437689b5ba673c595ba9c1040b9 Mon Sep 17 00:00:00 2001
+From: Anonymous Maarten <anonymous.maarten@gmail.com>
+Date: Wed, 15 Jul 2026 16:50:56 +0200
+Subject: [PATCH 02/12] bool is a C keyword from C23 onwards
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+This fixes the following error:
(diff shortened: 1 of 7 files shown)


sezero:
The following in yasm build stands out:

```
[78/127] Generating x86insn_gas.c
fail to map group of size 2 for tab size 512
[79/127] cmd.exe /c
```

But yasm build doesn't fail??

Is the issue related to python, e.g. python2 vs python3 ?

madebr:
> ```
> [78/127] Generating x86insn_gas.c
> fail to map group of size 2 for tab size 512
> ```
> Is the issue related to python, e.g. python2 vs python3 ?

I don't know but I don't think so. That error is emitted by C code ([`tools/genperf/perfect.c:565`](https://github.com/yasm/yasm/blob/a2f8bdf075bcad90a1b0a446080906ef6afe6847/tools/genperf/perfect.c#L565)).
It's a warning/error about failing to create a perfect hashing function.
I looked a bit into running the unit tests using CMake, but upstream's lack of feedback is discouraging.


**edit**: definitely no. It was already happening before:
https://github.com/madebr/yasm/actions/runs/27537918380/job/81391982415#step:9:596

madebr:
I cannot reproduce the error locally with MSVC 2019/2022/2026.
[The yasm unittests succeed ](https://github.com/madebr/yasm/actions/runs/29617241956) when using a MSVC toolchain.

madebr:
Upstream yasm can now build using latest CMake.

There's still a problem with the objects it generates though: [ci log](https://github.com/libsdl-org/SDL_mixer/actions/runs/29826755926/job/88696579600#step:12:634)
```
external\mpg123\ports\cmake\src\libmpg123\dct36_x86_64.S.obj : fatal error LNK1106: invalid file or disk full: cannot seek to 0x637
```

PR #867, opened 2026-06-16
[SDL2] cmake: Avoid misleading messages when smoke-testing library

smcv:
* cmake: Avoid misleading messages when smoke-testing library
    
    `Mix_Init(0)` doesn't attempt to initialize anything, so it's legitimate
    for it to return 0, and if it does, the error indicator won't be set,
    potentially leaving it with an unrelated (and benign) error message. Use
    a somewhat more realistic startup pattern: ask for an audio format that
    in practice is usually supported, and see whether it is.
    
    (If SDL_mixer is compiled without Ogg Vorbis support, that isn't an
    error, and this test will still succeed.)

---

Please cherry-pick to 2.8.x if accepted. This change is unnecessary for SDL3, which has an API change where `MIX_Init(void)` attempts to initialize all decoders that were supported at compile-time, without the user needing to specify which ones are potentially interesting.

### DIFF
--- a/cmake/test/main.c
+++ b/cmake/test/main.c
@@ -9,8 +9,10 @@ int main(int argc, char *argv[]) {
         fprintf(stderr, "SDL_Init: could not initialize SDL: %s\n", SDL_GetError());
         return 1;
     }
-    if (Mix_Init(0) == 0) {
-        fprintf(stderr, "Mix_Init: no sound/music loaders supported (%s)\n", Mix_GetError());
+    if (Mix_Init(MIX_INIT_OGG) & MIX_INIT_OGG) {
+        fprintf(stderr, "Mix_Init: Ogg Vorbis supported\n");
+    } else {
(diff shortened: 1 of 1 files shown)


smcv:
Windows CI failure seems unrelated:

> Compatibility with CMake < 3.5 has been removed from CMake

Looks like https://github.com/yasm/yasm/issues/285. Sorry, I'm not going to get into fixing that, but I see @madebr  is already on it.

madebr:
Yeah, yasm is generating broken objects on MSVC. Not sure yet what's wrong with it.

madebr (approved):

madebr:
> Please cherry-pick to 2.8.x if accepted. 

Cherry-picked as dafee306
I'm a bit surprised it worked. The release-2.x  branches of SDL are closed since https://discourse.libsdl.org/t/sdl-2-0-future-support/68372

smcv:
I thought the 2.x versions of the satellite projects (image, mixer, etc.) are still needed: it's only the SDL2 core library that has sdl2-compat as a (mostly) drop-in replacement.

Issue #868, opened 2026-06-17
Sounds play on repeat forever despite not specifying options when calling MIX_PlayTrack


sharkwouter:
So I'm working on upgrading my game to use SDL3 and for some reason this code is causing me issues:
```c
if (uneven_match) {
    MIX_StopTrack(this->channel_match1, 0);
    MIX_SetTrackGain(this->channel_match1, volume);
    MIX_SetTrackAudio(this->channel_match1, sound_match2);
    MIX_PlayTrack(this->channel_match1, 0);
} else {
    MIX_StopTrack(this->channel_match2, 0);
    MIX_SetTrackGain(this->channel_match2, volume);
    MIX_SetTrackAudio(this->channel_match2, sound_match1);
    MIX_PlayTrack(this->channel_match2, 0);
}
```

For some reason, these specific calls make the sound play on repeat forever.

I have almost the same code right above this code which looks like this and does not repeat:
```c
MIX_StopTrack(this->channel_pick, 0);
MIX_SetTrackGain(this->channel_pick, volume);
MIX_SetTrackAudio(this->channel_pick, sound_pick);
MIX_PlayTrack(this->channel_pick, 0);
```

Source: https://github.com/sharkwouter/oceanpop/blob/switch-to-SDL3/src/SoundManager.cpp

It is the switch-to-SDL3 branch. It is a pretty easy game to build with CMake, no options have to be specified. The libraries used are git submodules. I'm using SDL version 3.4.10 and SDL_mixer version 3.2.4.

Am I doing something wrong or am I running into a bug? I was able to reproduce this on MacOS, Linux, Windows and PSP.

icculus:
Sounds like a bug, I'll try to reproduce it over here.

sharkwouter:
Thanks a lot for looking inot this. So the way to reproduce it is to start a puzzle and create a match. Then you just do nothing and the matching sound will loop.

icculus:
Okay, I think this is reproducing here. So this sound repeating, right?

https://github.com/user-attachments/assets/1eede83d-ab39-4f9e-b5e4-ebc06bf859c2

icculus:
Okay, so it turns out match1.wav and match2.wav have a SMPL chunk that says the audio should loop, so the mixer is dutifully looping it forever when it plays that track.

You can tell SDL_mixer to ignore the looping metadata in the loaded audio, but it's a bit of a mouthful compared to just calling MIX_LoadAudio() ...

```diff 
[...]

PR #870, opened 2026-06-18
Set MIX_PROP_AUDIO_LOAD_IGNORE_LOOPS_BOOLEAN to true by default

sharkwouter:
I'm suggesting this change, because it recreates the old behavior that SDL2_mixer had. I doubt users will expect their audio to loop forever without telling SDL to do so. I also don't know of any media players that listen to this metadata and loops the audio. In short, I don't think the change made benefits users and this PR goes back to the old behavior.

### DIFF
--- a/include/SDL3_mixer/SDL_mixer.h
+++ b/include/SDL3_mixer/SDL_mixer.h
@@ -765,7 +765,7 @@ extern SDL_DECLSPEC MIX_Audio * SDLCALL MIX_LoadAudioNoCopy(MIX_Mixer *mixer, co
  *   the audio data specifying loop points. This will make a file decode from
  *   start to finish without looping, even if the file specified it should
  *   have. This audio can still be looped at playback time via MIX_Track loop
- *   settings, regardless of this setting. Default false.
+ *   settings, regardless of this setting. Default true.
  * - `MIX_PROP_AUDIO_DECODER_STRING`: the name of the decoder to use for this
  *   data. Optional. If not specified, SDL_mixer will examine the data and
  *   choose the best decoder. These names are the same returned from
--- a/src/decoder_drflac.c
+++ b/src/decoder_drflac.c
@@ -176,7 +176,7 @@ static bool SDLCALL DRFLAC_init_audio(SDL_IOStream *io, SDL_AudioSpec *spec, SDL
 
     adata->framesize = SDL_AUDIO_FRAMESIZE(*spec);
 
-    const bool ignore_loops = SDL_GetBooleanProperty(props, MIX_PROP_AUDIO_LOAD_IGNORE_LOOPS_BOOLEAN, false);
+    const bool ignore_loops = SDL_GetBooleanProperty(props, MIX_PROP_AUDIO_LOAD_IGNORE_LOOPS_BOOLEAN, true);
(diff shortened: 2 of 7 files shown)


slouken:
I'm not sure this is a good idea. The whole point of looping chunks is that they ... loop. :)

sharkwouter:
Up to you guys, I'd just want this change to be a deliberate choice if it stays around. Preferable we document it in the migration guide then, because for me it looked like a bug with no way to deal with it. Undocumented, I'd always assume it is a bug.

icculus:
So I don't think we should turn it off by default, but #868 showed me that _how_ you turn it off [kinda sucks](https://github.com/libsdl-org/SDL_mixer/issues/868#issuecomment-4738479836).

Maybe we need a blanket hint that says "don't ever loop anything" that apps can set with one line of code at startup and be done with it.

slouken:
> So I don't think we should turn it off by default, but #868 showed me that _how_ you turn it off [kinda sucks](https://github.com/libsdl-org/SDL_mixer/issues/868#issuecomment-4738479836).
> 
> Maybe we need a blanket hint that says "don't ever loop anything" that apps can set with one line of code at startup and be done with it.

This seems... incorrect. How often are people going to run into this, barring migrating unfamiliar assets from previous versions of SDL_mixer?

sharkwouter:
I would expect people to also sometimes want to use wav files they download which might have loop points set. They wouldn't expect the looping either, because the media player they preview the sound with will not loop it. To me it makes more sense to not loop automatically, but to use the loop points if the user enables looping explicitly, that is why I proposed this PR.

icculus:
> This seems... incorrect. How often are people going to run into this, barring migrating unfamiliar assets from previous versions of SDL_mixer?

I mean, @sharkwouter was surprised to find out his wav files have loop metadata. I suspect the universe of indie gamedev is driven by unfamiliar assets.

ligfx:
> I'm not sure this is a good idea. The whole point of looping chunks is that they ... loop. :)

I found the default behavior here confusing as well. IMO the `smpl` chunks are meant to be instructions on _how_ to loop when desired, not _to_ loop (based on their origination as instructions for MIDI samplers on how to keep an instrument's audio repeating when a key is held down) — though of course other people may have different expectations or be treating them differently in their asset collections.

icculus:
Hmm, that's a fair point for WAV files, if that's how they're still used (I honestly don't know)...but loop points in VOC, OGG, FLAC, and OPUS metadata are probably actually meant to be for looping the song for game background music purposes (specifically, so you have one intro part that plays once, and then the core song loops infinitely, or at least several times before a non-looping end segment), and I'd hesitate to remove that by default.

ligfx:
Yeah, I think it does makes sense to loop more "music"-oriented formats by default.

My personal ideal would be to remove the loop-overriding behavior from MIX_Audio entirely, and be able to control it entirely from MIX_Track. E.g., a MIX_Audio could expose default loop count and loop points that a MIX_Track picks up when I call MIX_SetTrackAudio/MIX_PlayTrack, but as a user I'd still like to be able to control the behavior using the same mechanisms as files without built-in loop data.

(Godot's an interesting example here: it defaults to "detect loops from file", but then lets the user explicitly disable or enable looping.)

PR #884, opened 2026-07-31
timidity soundfont support

### LINKED ISSUE #520: Soundfonts not supported by built-in Timidity?
I couldn't get MIDI sound to work in Chocolate Doom so I started doing some debugging.  I built SDL_mixer from source and can't even get a MIDI file to play with the sample `playmus` program.  It's just silence, even though that same file plays just fine with my system's Timidity++ (Arch Linux).

The only uncommented line in my timidity.cfg is `soundfont` + location.  I found this block of code which seems to imply that the soundfont line is completely ignored.  Am I reading this right?  It seems kind of insane that it's lacking this basic feature and I don't know how you are supposed to get sound without a soundfont.

https://github.com/libsdl-org/SDL_mixer/blob/142b884b67d226137298f56ab7e7e83b5580015d/src/codecs/timidity/timidity.c#L200-L210

sezero:
This is port of the work that recently went into libtimidity:

- Soundfont support, based on a timidity-0.2i patch by Takashi Iwai
  dating from 1997. Note that more than one soundfont directives in
  the config file is not supported: Only the first encountered font
  in timidity.cfg will be used.

- SDL_mixer decoder can reacts to a TIMIDITY_SOUNDFONT environment
  variable if available, signals embedded timidity to directly use
  the specified soundfont  and not parse timidity.cfg.

- Not ticking the template copyright checkbox: As noted above, almost
  all of the work is based on Takashi Iwai's old patch. In a private
  email to me, he agreed using his patch under Timidity dual license
  several years ago, because he follows the original licenses.
  And no, there is no AI involvement in this adaptation.

Please test and review.

Closes: https://github.com/libsdl-org/SDL_mixer/issues/520

P.S.: Posted an equivalent patchset to SDL_sound, too: https://github.com/icculus/SDL_sound/pull/126

- [ ] I confirm that I am the author of this code and release it to the SDL_mixer project under the Zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

### DIFF
--- a/src/timidity/instrum.c
+++ b/src/timidity/instrum.c
@@ -18,6 +18,7 @@
 #include "options.h"
 #include "common.h"
 #include "instrum.h"
+#include "sndfont.h"
 #include "resample.h"
 #include "tables.h"
 
@@ -208,6 +209,8 @@ static void load_instrument(MidiSong *song, const char *name,
   *out=SDL_malloc(sizeof(Instrument));
   ip = *out;
   if (!ip) goto nomem;
+
+  ip->type = INST_GUS;
 
   ip->samples = tmp[198];
   ip->sample = SDL_malloc(sizeof(Sample) * ip->samples);
@@ -543,6 +546,14 @@ static int fill_bank(MidiSong *song, int dr, int b)
 	    }
 	  else
 	    {
+	      /* preload soundfont */
+	      bank->instrument[i] = load_soundfont(song, 0,
+							(dr)? 128 : b,
+							(dr)? b : i,
+							(dr)? i : -1);
+	      if (bank->instrument[i])
+		continue;
+	      /* try gus patch */
 	      load_instrument(song,
 				     bank->tone[i].name,
 				     &bank->instrument[i],
@@ -559,6 +570,13 @@ static int fill_bank(MidiSong *song, int dr, int b)
 				     bank->tone[i].strip_envelope :
 				     ((dr) ? 1 : -1),
 				     bank->tone[i].strip_tail);
+	      if (bank->instrument[i])
+		continue;
+	      /* no patch; search soundfont again. */
+	      bank->instrument[i] = load_soundfont(song, 1,
+							(dr)? 128 : b,
+							(dr)? b : i,
+							(dr)? i : -1);
 	      if (!bank->instrument[i]) {
 		SNDDBG(("Couldn't load instrument %s (%s %d, program %d)\n",
 		   bank->tone[i].name,
--- a/src/timidity/instrum.c
+++ b/src/timidity/instrum.c
@@ -552,7 +552,7 @@ static int fill_bank(MidiSong *song, int dr, int b)
 							(dr)? b : i,
 							(dr)? i : -1);
 	      if (bank->instrument[i])
-		continue;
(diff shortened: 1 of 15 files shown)


src/timidity/sndfont.cfg
```diff
@@ -0,0 +1,19 @@
+#----------------------------------------------------------------
```
madebr, line 1: Users of SDL_mixer can configure the soundfont and cfg using the `TIMIDITY_SOUNDFONT` and `TIMIDITY_CFG` environment variables.
Would it make sense to turn these into SDL hints? These would still honor environment variables.

Those 2 variables are the only use of `SDL_getenv` in SDL_mixer. It appears hints are currently not used in SDL_mixer.

sezero: I followed the existing behavior. Your suggestion makes sense, but should be handled in a separate pull request, IMO. However, if `TIMIDITY_CFG` is changed in master, I can rebase and change `TIMIDITY_SOUNDFONT` accordingly.

madebr: I understand, I was just asking since this pr's topic is timidity.

`timidity/options.h` does `#define TIMIDITY_CFG "timidity.cfg"`, so it would make sense to name this file `timidity.cfg` instead.

sezero: >  name this file `timidity.cfg` instead.

You mean sndfont.cfg? Nah, it's just an example config from the original patch. Just leave it as is.

madebr: That's fine by me :)

src/timidity/instrum.c
```diff
@@ -543,6 +546,14 @@ static int fill_bank(MidiSong *song, int dr, int b)
 	    }
 	  else
 	    {
+	      /* preload soundfont */
+	      bank->instrument[i] = load_soundfont(song, 0,
+							(dr)? 128 : b,
+							(dr)? b : i,
+							(dr)? i : -1);
+	      if (bank->instrument[i])
+		continue;
```
slouken: spacing, here and below.

slouken (approved):
In general this change seems fine to me, but I'll leave it to @icculus to make the final decision. It might be worth running it through clang-format to get spacing and style consistent with the rest of the code.

sezero:
>  It might be worth running it through clang-format to get spacing and style consistent with the rest of the code.

I left the timidity code style as is and tried to stick to it when making changes in there. As far as I know, we have never changed timidity code style into ours, either. I'd leave such work to you guys.

src/timidity/instrum.c
sezero: Done.

icculus:
Yeah, I'd avoid reformatting it.

I think this is good, I'm going to merge it.

sezero:
> > This probably is not for the SDL2 branch.
> > For those interested, here is a patchset, backporting this to the SDL2 branch:
> > patchset-for-SDL2-branch.zip]
> > Unzip the attached archive, change into the SDL2 branch, and then run `git am 0*.patch` to apply.
> 
> 

https://github.com/libsdl-org/SDL_mixer/pull/897

PR #886, opened 2026-08-04
MIX_CreateTrack(): Set MIX_Track alignment to 16 instead of SDL_GetSIMDAlignment()

### LINKED ISSUE #885: MIX_Spatialize assertion failure on emscripten no matter what I do.
Image

I've tried everything I can to get rid of this. The only thing that ever fixes this is compiling in release mode. I even tried compiling SDL with assertions disabled and it still doesn't work.

Sackzement:
Fixes #885

### DIFF
--- a/src/SDL_mixer_internal.h
+++ b/src/SDL_mixer_internal.h
@@ -137,7 +137,7 @@ struct MIX_Audio
 
 struct MIX_Track
 {
-    float SDL_ALIGNED(16) position3d[4];   // we only need the X, Y, and Z coords, but the 4th element makes this SIMD-friendly.
+    float position3d[4];   // we only need the X, Y, and Z coords, but the 4th element makes this SIMD-friendly.
     MIX_SpatializationMode spatialization_mode;
     float spatialization_panning[2];
     int spatialization_speakers[2];
--- a/src/SDL_mixer_spatialization.c
+++ b/src/SDL_mixer_spatialization.c
@@ -410,7 +410,7 @@ void MIX_Spatialize(const MIX_VBAP2D *vbap2d, const float *position, float *pann
 {
     const int output_channels = vbap2d->speaker_count;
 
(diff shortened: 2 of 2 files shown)


icculus:
The assertion is wrong, not the alignment.

Sackzement:
I was under the impression that the assertion:
https://github.com/libsdl-org/SDL_mixer/blob/5a4422f112122b2b071bc11e14b5c22b5de9fd94/src/SDL_mixer_spatialization.c#L413

checks if the member alignment specifier is true:
https://github.com/libsdl-org/SDL_mixer/blob/5a4422f112122b2b071bc11e14b5c22b5de9fd94/src/SDL_mixer_internal.h#L140

But the allocation breaks this, because `SDL_GetSIMDAlignment()` can be smaller than 16:
https://github.com/libsdl-org/SDL_mixer/blob/5a4422f112122b2b071bc11e14b5c22b5de9fd94/src/SDL_mixer.c#L1435

If it is the assertion that is wrong, then the member alignment specifier should also change.

But the problem is that the member alignment specifier has to be a constant. The assertion and the allocation can be dynamic but should match the member alignment specifier.

icculus:
I think that SDL_ALIGNED() can probably go.

The 16 is aligning for SSE and NEON (which are the only codepaths we currently have, to be fair), but it's also not much use here. It's the first field in the struct, which I assume means that local allocations might use this to guarantee alignment when making space on the stack (which we never do). Maybe if it weren't, it might know to add padding if previous fields would cause misalignment.

But these things aren't relevant to our use-case, so we might as well remove it, fix the assertion, and keep aligning to system SIMD requirements, so this doesn't break when we add other SIMD instructions later.

Sackzement:
I updated the commit:
- Removed the member alignment.
- The assertion checks now for SDL_GetSIMDAlignment() instead of 16.
- The aligned allocation is back to SDL_GetSIMDAlignment().

icculus:
Thank you!

PR #895, opened 2026-08-26
stb_vorbis: zero-entry codebook support.

sezero:
Mainstream pull request: https://github.com/nothings/stb/pull/1990

Authored by: Alejandro González

@icculus, @slouken: Please review. If this is good and if we do accept this,
then I'll cherry-pick to all branches and also apply to SDL_sound.

( /cc @AliceLR for an additional pair of eyes and pıssisble adoption into
libxmp. )

( /cc @AlexTMjugador as the original author. )

Relevant parts of original commit message by the author is as follows.

```
Vorbis I specification [1] defines how to read and use codebooks with
several entries encoded in the setup header, and it also describes the
expected decoder behavior for the edge case of single-entry codebooks.
However, codebooks without entries are not clearly addressed.

It is my interpretation of the specification that zero-entry codebooks
are not forbidden, and are therefore allowed, provided that no bound on
the codebook entry count is violated by its use for residue or vector
decoding and, obviously, that the codebook is not actually used in any
audio packet.

While no Vorbis encoder I know of generates streams with zero-entry
codebooks, my OptiVorbis project recently added unused codebook entry
truncation support [2], which I verified to work successfully with the
reference libvorbis decoder. However, after that change had circulated
for a while, people began encountering issues playing files containing
such codebooks with both stb_vorbis and JOrbis.

The changes proposed here skip the problematic allocations during setup
header decoding when there are no entries, which have been tested to
allow stb_vorbis to work with several files containing zero-entry
codebooks.

[1] https://xiph.org/vorbis/doc/Vorbis_I_spec.html
[2] https://github.com/OptiVorbis/OptiVorbis/pull/187
```

- [ ] I confirm that I am the author of this code and release it to the SDL_mixer project under the Zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

### DIFF
--- a/src/stb_vorbis/stb_vorbis.h
+++ b/src/stb_vorbis/stb_vorbis.h
@@ -35,7 +35,7 @@
 //    github:audinowho   Dougall Johnson     David Reid
 //    github:Clownacy    Pedro J. Estebanez  Remi Verschelde
 //    AnthoFoxo          github:morlat       Gabriel Ravier
-//    Alice Rowan
+//    Alejandro González Alice Rowan
 //
 // Partial history:
 //    1.22    - 2021-07-11 - various small fixes
@@ -1042,7 +1042,7 @@ static void *setup_temp_malloc(vorb *f, uint64 siz)
       f->temp_offset -= sz;
       return (char *) f->alloc.alloc_buffer + f->temp_offset;
    }
-   return malloc(sz);
+   return sz ? malloc(sz) : NULL;
 }
 
 static void setup_temp_free(vorb *f, void **_p, uint64 sz)
@@ -3890,7 +3890,7 @@ static int start_decoder(vorb *f)
       uint32 *values;
       int ordered, sorted_count;
       int total=0;
-      uint8 *lengths;
+      uint8 *lengths=NULL;
       Codebook *c = f->codebooks+i;
       CHECK(f);
       x = get_bits(f, 8); if (x != 0x42)            return error(f, VORBIS_invalid_setup);
@@ -3907,13 +3907,15 @@ static int start_decoder(vorb *f)
       if (c->dimensions == 0 && c->entries != 0)    return error(f, VORBIS_invalid_setup);
       if (f->valid_bits < 0)                        return error(f, VORBIS_unexpected_eof);
 
-      if (c->sparse) {
-         lengths = (uint8 *) setup_temp_malloc(f, c->entries);
-         f->temp_lengths = lengths;
-      } else
-         lengths = c->codeword_lengths = (uint8 *) setup_malloc(f, c->entries);
+      if (c->entries > 0) {
+        if (c->sparse) {
+           lengths = (uint8 *) setup_temp_malloc(f, c->entries);
+           f->temp_lengths = lengths;
+        } else
+           lengths = c->codeword_lengths = (uint8 *) setup_malloc(f, c->entries);
 
-      if (!lengths) return error(f, VORBIS_outofmem);
+        if (!lengths) return error(f, VORBIS_outofmem);
+      }
 
       if (ordered) {
          int current_entry = 0;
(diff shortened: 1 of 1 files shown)


slouken (approved):

src/stb_vorbis/stb_vorbis.h
AliceLR, line 3996: I suspect taking the latter code path here on 0 entries is *not* what we want.

I will check the Vorbis spec to see if it says anything about 0 entries.

AlexTMjugador: For what it's worth, when writing this patch I thought that code path wasn't relevant because `sorted_entries` remained at 0, so the `if (c->sorted_entries) {` branch was skipped, and I couldn't tell that the initialization done in it was necessary later on. But please let me know if that's wrong, or if you spot anything I could use to improve the patch!

AliceLR: In addition to the specification saying nothing about 0-entry codebooks, [libvorbis passes codebook sizes of 0 directly to `malloc`](https://github.com/xiph/vorbis/blob/1b75110b5a2754ba1931d82dd83cb822b266a21d/lib/codebook.c#L169) when debug isn't enabled, which will cause [implementation defined](https://en.cppreference.com/c/memory/malloc) behavior. So, your 0 entry codebooks *probably* also fail for the reference implementation on some platforms (if I am remembering correctly, `malloc(0)` returns `NULL` with MinGW, MSVCRT, probably other libcs).

I recommend that you use single entry codebooks instead, which *are* [well-defined in the specification](https://www.xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-540003.2.1).

Otherwise, I think it would be better to not have non-sparse codebooks enter the sparse codebooks codepath e.g.:
```C
      if (!c->sparse) {
         if (c->entries > 0) {
            c->codewords = (uint32 *) setup_malloc(f, sizeof(c->codewords[0]) * c->entries);
            if (!c->codewords)                  return error(f, VORBIS_outofmem);
         }
      } else {
```

AliceLR:
@sezero I will fuzz this later (with the adjustment in the above conversation) just to verify `NULL` codeword arrays aren't somehow getting somewhere they aren't supposed to...

sezero:
OK, converted to draft, for now.

src/stb_vorbis/stb_vorbis.h
sezero: > (if I am remembering correctly, `malloc(0)` returns `NULL` with MinGW, MSVCRT, probably other libcs).

Don't know about others, but malloc() from openwatcom's libc does return NULL for zero-size params.

Unfortunately, SDL_malloc() retruns SDL_malloc(1) for zero-size params, after [a probably bad proposal of mine](https://github.com/libsdl-org/SDL/issues/2660), and [it remained unchanged in SDL3](https://github.com/libsdl-org/SDL/issues/6722) because @slouken had compatibility concerns.

AlexTMjugador: > In addition to the specification saying nothing about 0-entry codebooks [...]

Indeed, that was my finding as well. The algorithms it describes for reading codebook entries are, as far as I can tell, well-defined even when the number of entries is zero. So my conclusion was that they are technically valid, just like a codebook with 2<sup>20</sup> entries is also technically valid (and [a potential DoS vector I've thought about before](https://github.com/OptiVorbis/OptiVorbis/blob/07850e18d95fc9273059349eae700689189d459f/README.md#-known-limitations) :slightly_smiling_face:).

Your point about `libvorbis` having undefined behavior in this case is quite useful, though, because the tests I ran on `libvorbis` were done on Linux, where `libc` implementations [return a unique non-null pointer for `malloc(0)`](https://man7.org/linux/man-pages/man3/malloc.3.html), masking that problem.

So, given that the spec was not written with that case in mind, and that not even the reference implementation got it right, the "useful specification" to follow is that zero-entry codebooks are effectively disallowed because implementations rely on the behavior of the underlying allocators in this case. ([Protocol ossification](https://en.wikipedia.org/wiki/Protocol_ossification) is so lovely! :joy:)

> I recommend that you use single entry codebooks instead, which are [well-defined in the specification](https://www.xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-540003.2.1).

I agree that this the safer approach to take here. But as OptiVorbis is meant to produce more space-optimal setup headers for Vorbis streams, I think you can see how "wasting" a few bits to encode unused entries causes a little bit of tension. But I guess that's a me problem to solve, and ideally, one would remove unused codebooks from the header altogether... if they are not used by residue decoding or otherwise required by some other complication.

Failing that, I think it would make sense for OptiVorbis to add options to prevent it from performing aggressive optimizations like this that may not play well with deployed decoders, as risking the generation of streams that are effectively broken from a user's point of view is not a good default.

Regardless of the above adjustments to OptiVorbis, though, the can of worms of zero-entry codebooks has been opened, so I'll await your fuzzing results and adjust my patch accordingly. You clearly have more experience than I do with the `stb_vorbis` code in particular, so you'll hopefully be better placed to distinguish relevant fuzzer findings from noise.

Thanks for your comments!

sezero:
I am giving up on this one for the time being.

PR #897, opened 2026-09-02
[sdl2] timidity soundfont support

sezero:
SDL2 version of https://github.com/libsdl-org/SDL_mixer/pull/884


icculus:
I think mpg123 broke, unrelated to this PR, so I'm merging this.

